PR #30, opened 2024-06-02
feat: add `initialIndex` & `onMount` props

### LINKED ISSUE #28: Ability to open sheet on mount
Right now there is no real way to mount the sheet in opened state. We have to open it programmatically. 

A prop like `openOnMount` would be nice so the sheet opens automatically when mounted to. I personally want to use this so I can have a route that just have the TrueSheet and it opens when navigating to that screen.

Right now I can do a slight workaround like:

```
 const hasMounted = useRef(false);

  useEffect(() => {
    setTimeout(() => {
      if (hasMounted.current) return;
      hasMounted.current = true;
      handleOpen();
    }, 1000);
  }, []);
  ```
   Which works but does not feel very good

### PR DESCRIPTION
**Pull request by lodev09**
- Add `initialIndex` prop to present the sheet after loading.
- Add `onLoad` event prop to allow presenting sheet after loading.

Fixes #28

### CODE REVIEW

File: ios/TrueSheetView.swift [hunk 1]
```diff
@@ -275,6 +302,31 @@ class TrueSheetView: UIView, RCTInvalidating, TrueSheetViewControllerDelegate {
...
+    setContentHeight(NSNumber(value: contentHeight))
+
+    // Set initial footer height
+    let footerHeight = footerView.bounds.height
+    setFooterHeight(NSNumber(value: footerHeight))
+
+    // Present sheet at initial index
+    let initialIndex = self.initialIndex.intValue
+    if initialIndex >= 0 {
+      present(at: initialIndex, promise: nil, animated: false)
```
intergalacticspacehighway, line 324: Can we add [animateOnMount](https://ui.gorhom.dev/components/bottom-sheet/props#animateonmount) 🙏 . In our usecase we mount the sheet on navigation so it needs to be animated. Can be also done in a separate PR to reduce this PR's scope!

lodev09: This pr includes `onReady` event which you can take advantage to present the sheet on mount.

intergalacticspacehighway: yeah that works as well. Thanks

hirbod: Even though onReady is neat, I think an animateOnMount prop would be best. It would save one JS roundtrip: Event -> Callback -> Present -> Bridge -> Native react.

Having a prop and instantly opening on native whenever the sheet is ready should be the fastest solution imho.

lodev09: That makes sense. Agreed 👍

hirbod: Thanks. If initialIndex is only used for animateOnMount, we could just have one prop instead of to. Either pass a number to animateOnMount or just use initialIndex imo.

lodev09: I would prefer two props to make it more flexible. I would imagine some user to disable animation for the initial index, like on a maps app. `animateInitialIndex` would actually be more appropriate name for it.

hirbod: Just nits, but I don't like the naming of `animateInitialIndex` :D
Anyway, thanks for your work

hirbod (approved):
Hey @lodev09, I tested it thoroughly. It definitely feels snappier and faster now. Hopefully, we can get a faster easing function in a future release, but this works great for now. The auto height calculation also seems more accurate.

LGTM

Tested with and without animation, and also multiple stacks. I'll just need to adjust my paddings now, but thats fine, because its much more accurate and predictable now!

https://github.com/lodev09/react-native-true-sheet/assets/504909/adb1508e-116e-4bee-98ed-1f8a68b0d995

PR #37, opened 2024-06-10
fix: uiManager can be nil

hirbod:
I talked with Tomasz from Expo and he said that uiManager can be nil too, so it was fine in the first place.


lodev09:
Screenshot 2024-06-11 at 04 02 20

Not sure if `?` will help though... it could still crash if this becomes `nil`?

lodev09:
I'd say we stick to the type provided by RN. if it crashes, then there's definitely something else going on and might help dev to debug.

PR #70, opened 2024-07-19
chore: set IOS min version to 13.4 (RN 0.73)

### LINKED ISSUE #67: systemMaterialDark is only available in ios 13.0 or newer expo sdk49
i got this error when building ios app with expo sdk49 and true sheet `v0.12.3`

image

i already set my minimum deployment target to 13 and even tried above version, but still got the same error,

any idea why this is happen?

repo to reproduce:

https://github.com/bryanltobing/expo49-truesheet-error

steps:
- clone and yarn install repo
- npx expo prebuild -p ios
- npx expo run:ios

seems like the issue doesn't appear in expo 50 and above. but i need this to work in expo 49 since we are still unable to upgrade at the moment

lodev09:
Fix #67 

- Set minimum IOS version to `13.4` (RN 0.73 equivalent)

### DIFF
--- a/TrueSheet.podspec
+++ b/TrueSheet.podspec
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => min_ios_version_supported }
+  s.platforms    = { :ios => "13.4" }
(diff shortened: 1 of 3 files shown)


lodev09:
:rocket: _This pull request is included in v0.12.4. See [Release 0.12.4](https://github.com/lodev09/react-native-true-sheet/releases/tag/v0.12.4) for release notes._

PR #103, opened 2024-11-10
Edge-to-edge support

zoontek:
@lodev09 Hi 👋

Following https://github.com/lodev09/react-native-true-sheet/issues/72

I had to use `decorView.systemUiVisibility = View.SYSTEM_UI_FLAG_LAYOUT_STABLE or View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION` to ignore insets applied on the `BottomSheetDialog` internal `bottomSheet` `FrameLayout` ([related issue](https://github.com/material-components/material-components-android/issues/3389)). The alternative would be to fork and remove all the things we don't want, but…meh 😄.

Tested on Android 6 to 15

https://github.com/user-attachments/assets/629995a6-07c6-4736-9bd9-9f4f5294df60

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
@@ -22,6 +22,10 @@ class TrueSheetDialog(private val reactContext: ThemedReactContext, private val 
   private var sheetView: ViewGroup
   private var windowAnimation: Int = 0
 
+  override fun getEdgeToEdgeEnabled(): Boolean {
+    return if (Utils.EDGE_TO_EDGE) true else super.getEdgeToEdgeEnabled()
+  }
+
   /**
    * Specify whether the sheet background is dimmed.
    * Set to `false` to allow interaction with the background components.
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
@@ -22,10 +22,6 @@ class TrueSheetDialog(private val reactContext: ThemedReactContext, private val 
   private var sheetView: ViewGroup
   private var windowAnimation: Int = 0
 
-  override fun getEdgeToEdgeEnabled(): Boolean {
-    return if (Utils.EDGE_TO_EDGE) true else super.getEdgeToEdgeEnabled()
-  }
-
   /**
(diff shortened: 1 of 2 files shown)


lodev09:
Hey @zoontek, thank you so much for this!

If you don't mind, can we also add `EDGE_TO_EDGE` as a prop in the component as well?

zoontek:
@lodev09 , Sure, here it is. The prop value cannot be updated once the dialog is opened, or we would have to add a mechanism similar to RN Modal ([which dismiss and recreate the Dialog](https://github.com/facebook/react-native/blob/v0.76.1/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.kt#L223))

`react-native-is-edge-to-edge` has been added (it's super lightweight) for `edgeToEdge` prop default, similar to other libraries using it.

lodev09:
I think we can just put the burden to them on installing/detecting it 😂. So a simple prop would suffice. What do you think?

We can install both your lib in the example though, for demonstration purposes. I will update the docs in a separate PR.

Thanks again for this @zoontek

zoontek:
`react-native-is-edge-to-edge` is not `react-native-edge-to-edge`, it's a simple library without native code, just used for detection.

lodev09:
Okay, you do have a point. However, I'm one of those guys that tries to avoid unnecessary dependencies as much as possible 😅

Maybe you can revert it back to checking in the native side instead? If I'm not mistaken, this util checks if `react-native-edge-to-edge` is installed right?

So we check internally if `react-native-edge-to-edge` is installed AND also give them option to pass it as a prop.

What do you think?

zoontek:
@lodev09 `react-native-is-edge-to-edge` has another advantage compared to Kotlin check: if we continue and merge [this proposal](https://github.com/facebook/react-native/pull/47554), the package will be updated to detect both: first, if edge-to-edge is enabled in RN core, then if the user has installed `react-native-edge-to-edge`.

You will have nothing to do on your side, the transition will be smooth. Otherwise, you will need to perform a new release to handle both in Kotlin.

lodev09 (approved):

lodev09:
Looks good @zoontek. Thanks for your contribution 💪

lodev09:
:rocket: _This pull request is included in v1.0.0. See [Release 1.0.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v1.0.0) for release notes._

PR #108, opened 2024-11-21
fix(iOS): null checks in getTrueSheetView

dmnkgrc:
Similarly to https://github.com/lodev09/react-native-true-sheet/pull/35

This PR fixes a crash in onDismiss caused by nil values during app reloads in release builds. Both bridge and the view can be null and this is now handled in `getTrueSheetView`. In our case it looks like there is a race condition and `onDismiss` was called on a non existent sheet.

### DIFF
--- a/ios/TrueSheetViewManager.swift
+++ b/ios/TrueSheetViewManager.swift
@@ -24,9 +24,11 @@ class TrueSheetViewManager: RCTViewManager {
 
   // MARK: - Private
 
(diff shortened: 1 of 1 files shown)


lodev09 (approved):

lodev09:
Thank you @dmnkgrc

lodev09:
:rocket: _This pull request is included in v1.0.0. See [Release 1.0.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v1.0.0) for release notes._

PR #117, opened 2025-01-22
fix(android): support react native 0.77

believer:
Thanks for this great library!

Due to [this change in React Native core](https://github.com/facebook/react-native/commit/145c72f8163048f0eee30d5ce850911f5dad865c#diff-cab6c32d3e05344091c804ea93e8a220dabdd75ec2a258aaf4bde7751c13f844R31) `getString` now returns `String?`. Since the `when` condition checks that the current value is a string, I think it's fine to type cast the result as a `String`. I tested it locally and it resolves the error.

Alternatively, we could go the route of only adding the value if it exists using:

```kotlin
	ReadableType.String -> sizes.getString(i)?.let { result.add(it) }
```

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
@@ -107,6 +107,9 @@ class TrueSheetViewManager : ViewGroupManager<TrueSheetView>() {
     for (i in 0 until minOf(sizes.size(), 3)) {
       when (sizes.getType(i)) {
         ReadableType.Number -> result.add(sizes.getDouble(i))
+        // React Native < 0.77 used String for getString, but 0.77
+        // changed it to String?. Suppress the error for older APIs.
+        @Suppress("UNNECESSARY_SAFE_CALL")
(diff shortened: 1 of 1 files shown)


believer:
@lodev09 Hi! What do you think about these changes to support React Native 0.77?

lodev09:
Hey @believer, thanks for the PR. I would prefer null check instead of casting. I'm not sure if this will cause compile issue on older RN version tho.

Thanks!

believer:
Changed it to a null check instead, I agree that it's cleaner. I doesn't seem to cause a compile error, but it displays a warning:

```
Unnecessary safe call on a non-null receiver of type 'kotlin. String'.
```

lodev09:
Yep, suppress annotation should be fine! Thanks @believer

believer:
Awesome, will fix it first thing on Monday!

lodev09 (approved):
LGTM. Thank you!

lodev09:
:rocket: _This pull request is included in v1.1.0. See [Release 1.1.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v1.1.0) for release notes._

PR #121, opened 2025-01-26
feat: backgroundColor apply refactoring

sergeymild:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lodev09/react-native-true-sheet",
-  "version": "1.0.3",
+  "version": "1.0.4",
```
lodev09, line 3: Can you revert this please? We'll let the release automatically update this.

src/TrueSheet.tsx
```diff
@@ -250,7 +251,7 @@ export class TrueSheet extends PureComponent<TrueSheetProps, TrueSheetState> {
         scrollableHandle={this.state.scrollableHandle}
         sizes={sizes}
         blurTint={blurTint}
-        background={backgroundColor}
+        background={(backgroundColor ? processColor(backgroundColor) : undefined) as any}
```
lodev09, line 254: Maybe change the `background` type in `TrueSheetNativeViewProps` instead of casting to `any`

lodev09 (commented):
Looks good. See minor comments

src/TrueSheet.tsx
sergeymild: If you change this property in TrueSheetNativeViewProps so that its type is number instead of ColorType, the user won't be able to use it as a regular color property. I don't see any problem with casting to any because it's part of the internal implementation and won't be exposed externally.

package.json
sergeymild: done

src/TrueSheet.tsx
lodev09: the `background` prop is actually just a prop for the native component. The user facing prop is `backgroundColor`.

lodev09, line 254:
```suggestion
        background={(backgroundColor ? processColor(backgroundColor) : undefined)}
```

src/TrueSheet.tsx
```diff
@@ -9,6 +9,7 @@ import {
   type NativeSyntheticEvent,
   type LayoutChangeEvent,
   type ColorValue,
+  processColor,
```
lodev09, line 12:
```suggestion
  type ProcessedColorValue,
```

lodev09: Replace the type with this:
```
interface TrueSheetNativeViewProps
  extends Omit<TrueSheetProps, 'onPresent' | 'onSizeChange' | 'backgroundColor'> {
  contentHeight?: number
  footerHeight?: number
  background?: ProcessedColorValue | null
  scrollableHandle: number | null
  onPresent: (event: SizeChangeEvent) => void
  onSizeChange: (event: SizeChangeEvent) => void
  onContainerSizeChange: (event: ContainerSizeChangeEvent) => void
}
```

lodev09 (changes requested):

lodev09 (approved):
going to merge this and update suggested code in main.
thanks @sergeymild

lodev09:
:rocket: _This pull request is included in v1.1.0. See [Release 1.1.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v1.1.0) for release notes._

Issue #84, opened 2024-08-18
Is there a way to get real-time SizeInfo when sliding the BottomSheet?


Richard6536:
Hi,

I’m using TrueSheet in my React Native project and I’m trying to get real-time updates on the SizeInfo (similar to what onSizeChange provides) while I’m sliding the BottomSheet. Currently, onSizeChange only triggers once the sliding gesture is completed. Is there any existing method or workaround that would allow me to get the SizeInfo continuously during the slide?

Thanks!

lodev09:
Yep.. It will come soon, just didn't get the chance to work on it yet -- should also support reanimated :)

Richard6536:
> Yep.. It will come soon, just didn't get the chance to work on it yet -- should also support reanimated :)

Thank you! I'm looking forward to it. Appreciate your hard work on this!

erickreutz:
This would be really helpful to have!

lodev09:
@erickreutz thanks. I was about to implement this but lost track. reopening

lodev09:
I'm going to take a look at some of new arch bugs first then will do this next. Thanks for the sponsor @erickreutz !

erickreutz:
> I'm going to take a look at some of new arch bugs first then will do this next. Thanks for the sponsor [@erickreutz](https://github.com/erickreutz) !

@lodev09 sounds good thank you!

PR #124, opened 2025-01-26
Feat: Drag events

lodev09:
This PR introduces drag events.
See #84 

reanimated


erickreutz:
A question we this - how would we be able to sync this with a `SharedValue` from reanimated? Something similar to `@gorhom/bottom-sheets` `animatedIndex` and `animatedPosition` props would be great.

lodev09:
Hi @erickreutz. I believe Reanimated supports this. Like [this one](https://github.com/callstack/react-native-pager-view?tab=readme-ov-file#reanimated-onpagescroll-handler) for example

erickreutz:
Ahh I see, yes that makes sense

lodev09:
@erickreutz drag events are ready for both IOS and android. However, I did some digging into integrating Reanimated and found out that I need to change the way the native dispatches these events so reanimated can support it.

See this [old comment](https://github.com/software-mansion/react-native-reanimated/issues/1229#issuecomment-735609106) for reference.

erickreutz:
@lodev09 lame! you'd think they would have implemented that in the last 4 years

lodev09:
@erickreutz I know right 😄 . I will adjust and grab code from pager-view for dispatching events. Pager view with reanimated works on the current App I'm working on.

erickreutz:
@lodev09 yeah we use that pager view with `useEvent` in our app as well.

lodev09:
@erickreutz can you give it a quick review? I tried to make it as simple as possible as there are some limitations on both ios & android :/

erickreutz:
@lodev09 looks good! Interested to see how this feels in our app

Issue #78, opened 2024-08-02
Diming intensity


ljukas:
I would like to be able to control the intensity of the dimming effect when the sheet is shown.

Right now its very lightly dimmed only, I would like to be able to make this darker to further enhance that attention should be directed to the sheet that is shown.

Here is a picture of a comparison between my usage of gorhom bottom sheet where I can control this, and true-sheet where I cannot control this.

![image](https://github.com/user-attachments/assets/a7412813-b589-4d99-ab81-95ef64673804)
![image](https://github.com/user-attachments/assets/34779b08-b001-4a39-aaa1-4532bae8413b)

Really nice lib this :D Thanks for all the work you do

coder966:
I believe even if PR #151 gets merged, the default intensity should be increased. What do you think @lodev09 ?

lodev09:
v3 is out! Added first-class support for Reanimated. You can create custom backdrop to do this. See this [guide](https://sheet.lodev09.com/guides/dimming/#customizing-dimming-alpha)

PR #151, opened 2025-02-26
feat: add `dimmedAlpha` prop on iOS

camchis:
* Allow customisable dim alpha for the root view on iOS through `dimmedAlpha` prop
* Supports stacked sheets
* Respects `dimmed` prop

Requested in #78 and also find it useful for my own use case

Happy to add to docs if this looks good

https://github.com/user-attachments/assets/e851c7c9-be21-490c-958f-9fd7db60b077

### DIFF
--- a/ios/TrueSheetView.swift
+++ b/ios/TrueSheetView.swift
@@ -211,19 +211,22 @@ class TrueSheetView: UIView, RCTInvalidating, TrueSheetViewControllerDelegate {
   }
 
   func viewControllerDidAppear() {
-    // Add this sheet's opacity to the stack
-    TrueSheetView.sheetOpacityStack.append(dimmedAlpha)
-
-    // Dim root view controller
-    UIView.animate(withDuration: 0.3) {
-      if let rootViewController = UIApplication.shared.windows.first?.rootViewController {
-        rootViewController.view.alpha = self.dimmedAlpha
+    // Only apply dimming if the sheet has dimmed property set to true
+    if viewController.dimmed {
+      // Add this sheet's opacity to the stack
+      TrueSheetView.sheetOpacityStack.append(dimmedAlpha)
+      
+      // Dim root view controller
+      UIView.animate(withDuration: 0.3) {
+        if let rootViewController = UIApplication.shared.windows.first?.rootViewController {
+          rootViewController.view.alpha = self.dimmedAlpha
+        }
       }
     }
   }
 
   func viewControllerWillDisappear() {
-    if viewController.isBeingDismissed {
+    if viewController.isBeingDismissed && viewController.dimmed {
       // Remove the topmost opacity value (which should be this sheet's)
       if !TrueSheetView.sheetOpacityStack.isEmpty {
         TrueSheetView.sheetOpacityStack.removeLast()
--- a/ios/TrueSheetView.swift
+++ b/ios/TrueSheetView.swift
@@ -8,6 +8,11 @@
 
 @objc(TrueSheetView)
 class TrueSheetView: UIView, RCTInvalidating, TrueSheetViewControllerDelegate {
+  // MARK: - Static properties
+
+  // Keep track of active sheet opacity values (stack)
+  private static var sheetOpacityStack: [CGFloat] = []
+
   // MARK: - React properties
 
   // MARK: - Events
@@ -203,17 +208,38 @@ class TrueSheetView: UIView, RCTInvalidating, TrueSheetViewControllerDelegate {
     // Add constraints to fix weirdness and support ScrollView
     contentView.pinTo(view: containerView, constraints: nil)
     scrollView.pinTo(view: contentView, constraints: nil)
+  }
 
-    if viewController.dimmed {
-      UIView.animate(withDuration: 0.3) {
-        self.viewController.presentingViewController?.view.alpha = self.dimmedAlpha
+  func viewControllerDidAppear() {
+    // Add this sheet's opacity to the stack
+    TrueSheetView.sheetOpacityStack.append(dimmedAlpha)
+
+    // Dim root view controller
+    UIView.animate(withDuration: 0.3) {
+      if let rootViewController = UIApplication.shared.windows.first?.rootViewController {
+        rootViewController.view.alpha = self.dimmedAlpha
       }
     }
   }
 
   func viewControllerWillDisappear() {
-    UIView.animate(withDuration: 0.3) {
-      self.viewController.presentingViewController?.view.alpha = 1
+    if viewController.isBeingDismissed {
+      // Remove the topmost opacity value (which should be this sheet's)
+      if !TrueSheetView.sheetOpacityStack.isEmpty {
+        TrueSheetView.sheetOpacityStack.removeLast()
+      }
+
+      UIView.animate(withDuration: 0.3) {
+        if let rootViewController = UIApplication.shared.windows.first?.rootViewController {
+          // If there are still active sheets, apply the alpha of the topmost one
+          if let topmostOpacity = TrueSheetView.sheetOpacityStack.last {
+            rootViewController.view.alpha = topmostOpacity
+          } else {
+            // If no sheets are left, restore alpha to 1
+            rootViewController.view.alpha = 1
+          }
+        }
+      }
     }
   }
 
--- a/ios/TrueSheetView.swift
+++ b/ios/TrueSheetView.swift
@@ -25,6 +25,7 @@ class TrueSheetView: UIView, RCTInvalidating, TrueSheetViewControllerDelegate {
 
   @objc var initialIndex: NSNumber = -1
   @objc var initialIndexAnimated = true
+  @objc var dimmedAlpha: CGFloat = 0.75
 
   // MARK: - Private properties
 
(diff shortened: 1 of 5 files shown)


lodev09:
Hey @camchis, thank you for this PR! I'll take a look tomorrow.

lodev09:
Hey @camchis, is it possible for you to add this feature on android as well?

camchis:
> Hey @camchis, is it possible for you to add this feature on android as well?

Will look into this today :)

camchis:
hey @lodev09, I've added Android support

camchis:
@lodev09 do you have any thoughts on this PR? Would be great to get it released 🙏

crux153:
@camchis It seems that the current code behaves oppositely on iOS and Android. If dimmedAlpha represents the alpha value, a lower number should make it more transparent, and a higher number should make it darker. This works as expected on Android, but on iOS, it's reversed—the lower the value, the darker it gets. To make it consistent, we should modify iOS to use `1 - dimmedAlpha`.

Also, on iOS, the dimming animation starts after the sheet is fully opened, which feels unnatural. Moving it to viewWillAppear would make it start while the sheet is opening, which looks more natural.

Here’s the updated code with these fixes.

```swift
func viewControllerWillAppear() {
  // Only apply dimming if the sheet has dimmed property set to true
  if viewController.dimmed {
    let opacity = 1 - dimmedAlpha;
    // Add this sheet's opacity to the stack and dim root view controller
    TrueSheetView.sheetOpacityStack.append(opacity)
    UIView.animate(withDuration: 0.3) {
      if let rootViewController = UIApplication.shared.windows.first?.rootViewController {
        rootViewController.view.alpha = opacity
      }
    }
  }

  guard let contentView, let scrollView, let containerView else {
    return
  }

  // Add constraints to fix weirdness and support ScrollView
  contentView.pinTo(view: containerView, constraints: nil)
  scrollView.pinTo(view: contentView, constraints: nil)
}

func viewControllerDidAppear() {
}
```

camchis:
Thanks @crux153!!
Both very good points

I've pushed a change to add the dimming logic inside `viewControllerWillAppear` and have removed `viewControllerDidAppear`

Have also made the `dimmedAlpha` behaviour consistent with Android, good catch :)

crux153:
@camchis Great work!

And for Android implementation, using the built-in `setDimAmount` method seems simpler than using a custom view.
You can just add one line `setDimAmount(dimmedAlpha)` to the `setupDimmedBackground` method in TrueSheetDialog.kt

```kotlin
/**
 * Setup dimmed sheet.
 * `dimmedIndex` will further customize the dimming behavior.
 */
fun setupDimmedBackground(sizeIndex: Int) {
  window?.apply {
    val view = findViewById<View>(com.google.android.material.R.id.touch_outside)

    if (dimmed && sizeIndex >= dimmedIndex) {
      // Remove touch listener
      view.setOnTouchListener(null)

      // Add the dimmed background
      setFlags(
        WindowManager.LayoutParams.FLAG_DIM_BEHIND,
        WindowManager.LayoutParams.FLAG_DIM_BEHIND
      )

      setDimAmount(dimmedAlpha)

      setCanceledOnTouchOutside(dismissible)
    } else {
      // Override the background touch and pass it to the components outside
      view.setOnTouchListener { v, event ->
        event.setLocation(event.rawX - v.x, event.rawY - v.y)
        reactContext.currentActivity?.dispatchTouchEvent(event)
        false
      }

      // Remove the dimmed background
      clearFlags(WindowManager.LayoutParams.FLAG_DIM_BEHIND)

      setCanceledOnTouchOutside(false)
    }
  }
}
```

camchis:
@crux153 thanks, I 100% agree that the `setDimAmount` method is a lot simpler than using a custom view, however I can't seem to get it working at all when I implement it, I'm struggling to figure out why which is why I went with a custom view in the end

crux153:
@camchis Here's the patch that I'm currently using.

```diff
diff --git a/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt b/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
index f4b38fc94de83678124a86bc3c7ab3e855e585b0..fddd0d46fb673669a093ba0b2faf27825391e761 100644
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
@@ -46,6 +46,13 @@ class TrueSheetDialog(private val reactContext: ThemedReactContext, private val
    */
   var dimmedIndex = 0
 
+  /**
+   * The alpha value of the dimmed background.
+   *
+   * @default 0.75f
+   */
+  var dimmedAlpha = 0.75f
+
   /**
    * The maximum window height
    */
@@ -156,6 +163,8 @@ class TrueSheetDialog(private val reactContext: ThemedReactContext, private val
           WindowManager.LayoutParams.FLAG_DIM_BEHIND
         )
 
+        setDimAmount(dimmedAlpha)
+
         setCanceledOnTouchOutside(dismissible)
       } else {
         // Override the background touch and pass it to the components outside
diff --git a/android/src/main/java/com/lodev09/truesheet/TrueSheetView.kt b/android/src/main/java/com/lodev09/truesheet/TrueSheetView.kt
index e946d96815bb49cb6f17a71587ed2fc4925e0625..9fd80c68ac7cc970d4cb4bb4d13bbd0e738cbbba 100644
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetView.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetView.kt
@@ -362,6 +362,12 @@ class TrueSheetView(context: Context) :
     }
   }
 
+  fun setDimmedAlpha(alpha: Float) {
+    if (sheetDialog.dimmedAlpha == alpha) return
+
+    sheetDialog.dimmedAlpha = alpha
+  }
+
   fun setDimmedIndex(index: Int) {
     if (sheetDialog.dimmedIndex == index) return
 
diff --git a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
index 0ed9b8a0e69cc529d421e34ef937186dfe7e72d0..edecc9570fb2a15d336212f9d49f02674043dcca 100644
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
@@ -67,6 +67,11 @@ class TrueSheetViewManager : ViewGroupManager<TrueSheetView>() {
     view.setDimmed(dimmed)
   }
 
+  @ReactProp(name = "dimmedAlpha")
+  fun setDimmedAlpha(view: TrueSheetView, alpha: Float) {
+    view.setDimmedAlpha(alpha)
+  }
+
   @ReactProp(name = "initialIndex")
   fun setInitialIndex(view: TrueSheetView, index: Int) {
     view.initialIndex = index
```

andrew-stupchuk:
Hey @lodev09. Could you please take a look at the grabber in the attached video? When it opens, it initially appears black, then quickly turns grey. Also, both bottom corners briefly flash white. I was wondering if there might be a workaround or fix for this? This issue is consistently reproducible when using the light theme

lxup:
Really excited to get the stacked sheet, it's the only thing I'm missing!

camchis:
@crux153, `setDimAmount()` is still not working for me with your patch and I'm unsure why - would be interested to know if it working for others.

adelbeke:
The improvement looks awesome 🤩

ndoherty-xyz:
Would love to get this merged in, it's exactly what I need for my use case.

lodev09:
Closing this. No longer needed in native. I wanted to retain the native defaults.

To support this, I've implemented first class support for Reanimated that you can use to add custom backdrop with custom alpha. This is now available on v3! Please test it out 🙏

lodev09:
Here's a guide: https://sheet.lodev09.com/guides/dimming/#customizing-dimming-alpha

camchis:
Nice work!

kristian-nst:
this is a lot of work just because we can't have 1 tiny prop?

it works, but a bit ridiculous just because we cant change '0.75' to another value :P 

```
function CustomBackdrop({
  sheetRef,
  useTransparentBackdrop,
}: {
  sheetRef: RefObject<TrueSheet>
  useTransparentBackdrop: boolean
}): ReactElement {
  const { animatedIndex } = useReanimatedTrueSheet()

  const targetOpacity = useMemo(() => {
    return useTransparentBackdrop ? 0.1 : 0.5
  }, [useTransparentBackdrop])

  const backdropStyle = useAnimatedStyle(() => {
    return {
      opacity: interpolate(
        animatedIndex.value,
        [-1, 0],
        [0, targetOpacity],
        Extrapolation.CLAMP,
      ),
    }
  })

  return (
    <Animated.View
      style={[
        {
          zIndex: 1000,
          position: 'absolute',
          top: 0,
          left: 0,
          right: 0,
          bottom: 0,
          backgroundColor: 'black',
        },
        backdropStyle,
      ]}
    >
      <Pressable
        onPress={() => {
          sheetRef.current?.dismiss()
        }}
        className="absolute bottom-0 left-0 right-0 top-0"
      />
    </Animated.View>
  )
}
```

also, seems like the custom one doesn't handle dismissals, is that on purpose? (missing in docs)

lodev09:
@kristian-nst I'll gladly add a prop if IOS supports this officially, but it's a lot of work and hack to get it to work on IOS so I'd rather leave it as-is for now 😄 

Maybe IOS 30 will have this option LOL

Issue #141, opened 2025-02-24
Question: Preventing sheet from closing when interacting with elements (Android)


szheleshchenko:
Hey, I ran into an issue on Android where the sheet closes when interacting with elements inside it (like a `DatePicker`, but could be any interactive element). On iOS, the sheet stays open when using

```
<View
  onStartShouldSetResponder={() => true}
  onResponderTerminationRequest={() => false}
/>
```

but on Android, these props don't work.

I tried to workaround it with adding a `ScrollView` and passing `scrollRef` to the sheet, but that was no use.

Is there a way to prevent this behavior on Android without disabling the sheet's movement altogether? It's a pretty common use case to have interactive elements inside the bottom sheet that shouldn’t trigger a close action when touched.

Would love some help or suggestions on how to handle this. Thanks! 🙌

lodev09:
Hey @szheleshchenko, do you have a reproducible example?

szheleshchenko:
@lodev09 Yes, I created an example using DatePicker. https://github.com/szheleshchenko/true-sheet-reproduction/blob/master/app/index.tsx

Here's how it looks:
<details open><summary>Details</summary>

https://github.com/user-attachments/assets/baf615a5-0d3c-420d-bc6c-2af7541d9584

</details>

skaionedev:
Any updates or solutions to fix this issue on android?

szheleshchenko:
> Any updates or solutions to fix this issue on android?

I use react-native-actions-sheet for this particular case. This library is practically unusable on iOS and has poor performance, but as a temporary solution for Android, it's fine.

lodev09:
@skaionedev your issue just needs `flexGrow` style on the `GestureHandlerRootView` wrapper.

@szheleshchenko should be fixed in PR #156 along with other android touch issues.

> This library is practically unusable on iOS and has poor performance

Can you open a separate issue regarding this? Thanks

szheleshchenko:
@lodev09 Sorry for the confusion. By 'this library,' I meant `react-native-actions-sheet`. Everything is fine with true-sheet on iOS. 🙂

lodev09:
@szheleshchenko ohh I see. no worries. Can you checkout fix in #156 please? :)

skaionedev:
@lodev09  `flexGrow` helps to enable dragging on Android but it doesn't solve my problem that I have with dragging top item to bottom  (bottom-sheet closes and prevents list from reordering) 
[...]

Issue #135, opened 2025-02-12
Pressable Items Not Responding Inside TrueSheet on Android


fukuli053:
### Environment
- Expo SDK: ~52.0.33
- Platform: Android
- New Architecture: Enabled

### Description
I'm encountering an issue where `Pressable` or `TouchableOpacity` components inside `TrueSheet` are not responding to click events on Android. This appears to be specifically related to the New Architecture being enabled. I also tried this with `GestureHandlerRootView ` but It did not work.

### Expected Behavior
- `Pressable` or `TouchableOpacity` components should respond to click events inside `TrueSheet`
- `dismiss()` function should close the sheet when called
- Console logs should be triggered on press events

### Actual Behavior
- `Pressable` or `TouchableOpacity` components do not respond to clicks inside `TrueSheet`

### Reproduction Steps

```tsx
import { Button, View } from 'react-native';
import { TrueSheet } from 'react-native-true-sheet';
import { useRef } from 'react';

export default function App() {
  const sheet = useRef(null);

  // Present the sheet
  const present = async () => {
    await sheet.current?.present();
    console.log('Sheet has been presented');
  };

  // Dismiss the sheet
  const dismiss = async () => {
    await sheet.current?.dismiss();
    console.log('Sheet dismissed');
  };

  return (
    <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
      <Button title="Open Sheet" onPress={present} />
      
      <TrueSheet ref={sheet}>
        <View style={{ padding: 20 }}>
          <Pressable 
            onPress={dismiss}
            style={({ pressed }) => [
              {
                backgroundColor: pressed ? '#ddd' : '#fff',
                padding: 15,
                borderRadius: 8,
              },
            ]}
          >
            <Text>Close Sheet</Text>
          </Pressable>
        </View>
      </TrueSheet>
    </View>
  );
}
```

lodev09:
@fukuli053 what version of TrueSheet are you on?

fukuli053:
> [@fukuli053](https://github.com/fukuli053) what version of TrueSheet are you on?

1.1.1

ozgurbayram:
I get this bug if new arch is true @lodev09 it works fine if new arch is false

nicolascavallin:
I'm facing same issues with Android. At the moment iOS is working. I'm not sure how to help

418442040:
I'm facing same issues with Android. At the moment iOS is working

[...]

PR #156, opened 2025-03-04
fix(android): touch events on android new arch

### LINKED ISSUE #154: Android New Architecture no press event.

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

So bit of a weird issue when running the latest version 2.0.3 and the new architecture on android, the onPress events of children seem to be getting swallowed.  I see that the interaction is happening with opacity changing but then no event is being sent to JS side meaning the onPress doesn't get triggered. 

Not 100% sure if another library is conflicting I am running through the android code now but just though i'd raise this bug incase anyone else had encountered this.

Update: Holding the button with one finger and pressing with another seems to enable the event callback.

### Library version

2.0.3

### Environment info

    "react-native": "^0.77.1",

### Steps to reproduce

..

### Reproducible example repository

N/A

lodev09:
Attempts to fix #154, #135 and #141 🙏


lodev09:
:rocket: _This pull request is included in v2.0.4. See [Release 2.0.4](https://github.com/lodev09/react-native-true-sheet/releases/tag/v2.0.4) for release notes._

rickyzhangca:
unfortunately issue still happens on physical devices

PR #192, opened 2025-07-01
Make android sheet reactive to height changes

adelbeke:
This PR fixes an issue with dynamic content height updates when using the 'auto' size option on Android.

It fixes the #191 issue.

## Issue
When using a TrueSheet with `sizes={['auto']}` and dynamically updating the content (e.g., conditional rendering after a delay), the sheet wouldn't resize properly on Android, even though it worked correctly on iOS.

## Solution
Enhanced the content height handling in `TrueSheetDialog.kt` to properly update the sheet's dimensions when content changes:
1. Added a setter for `contentHeight` that detects changes
2. When using 'auto' size, directly updates the container's layout params
3. Forces the behavior to reconfigure with the new height
4. Ensures the sheet expands to the new height by setting the proper state

## Testing
To test this fix:
1. Create a sheet with `sizes={['auto']}`
2. Add conditional content that renders after a delay
3. Verify the sheet properly resizes when the content appears

Before this fix, the sheet would remain at its initial height. After the fix, it properly expands to fit the new content.

## Preview

**Before**

<video src="https://github.com/user-attachments/assets/19246b54-3973-469b-8eb1-851802363ea5" width="200"></video>

**After**

<video src="https://github.com/user-attachments/assets/2880a358-c643-4cd3-a3a5-7bf5349a037a" width="200"></video>

PR #193, opened 2025-07-02
fix: android auto responsive sheet size

lovegaoshi:
this PR implements the strategy mentioned in https://github.com/lodev09/react-native-true-sheet/pull/192#issuecomment-3025669437 to auto adjust bottom sheet height on content height change. I replaced when the only size is "auto"'s bottomsheetbehavior to COLLAPSED instead of EXPANDED, so its height can be adjusted after the sheet is shown.

test is done using a custom component in my app instead of the test code mentioned in the above PR/linked issue, but the idea is the same.

https://github.com/user-attachments/assets/5a00981e-87e3-49a9-a004-0c935ff94437

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetDialog.kt
@@ -52,6 +52,20 @@ class TrueSheetDialog(private val reactContext: ThemedReactContext, private val 
   var maxScreenHeight = 0
 
   var contentHeight = 0
+    set(value) {
+      val oldValue = field
+      field = value
+
+      // If height changed and using auto size, reconfigure
+      if (oldValue != value && sizes.size == 1 && sizes[0] == "auto" && isShowing) {
+        // Force expanded state to ensure proper height
+
+        behavior.apply {
+          setPeekHeight(value + footerHeight, true)
+          state = BottomSheetBehavior.STATE_COLLAPSED
+        }
+      }
+    }
   var footerHeight = 0
   var maxSheetHeight: Int? = null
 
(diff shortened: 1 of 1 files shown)


lodev09:
Thanks @lovegaoshi. I'll test over the weekend.

lodev09:
@lovegaoshi tested and works. however, it is now draggable to full screen height. Can we disable that and have it draggable up until specified max height?

lovegaoshi:
I dont know if that is technically feasible - that requires changing maxHeight which per documentation should only be specified when the sheet is closed. Perhaps there are hacks that either 
- disable bottomsheet.behavior.EXPANDED
- or combine this with the setMaxHeight approach from https://github.com/lodev09/react-native-true-sheet/pull/192 on content size change hoping as bottomsheet.behavior is still in COLLAPSED it wont trigger layout changes.
but anyways i suspect hacks will be involved.

lodev09:
Yeah, try 2nd option. It might be advised to only set maxHeight when closed due for UX purposes since it can't be animated (bad UI for the user). Since we don't get any error so I guess it's kinda "ok" 😅

lovegaoshi:
ok. i actually got it working but its very hacked to me. 

we already saw https://github.com/lodev09/react-native-true-sheet/pull/192 doesnt give you an animation. the issue is bottomSheetBehavior.state === EXPANDED will NOT animate. only COLLAPSED with seekHeight animates. 

actually if instead of using EXPANDED in #192 but use COLLAPSED, and put bottomSheet.state = COLLAPSED **before** setting the bottomsheetView.layout, this will work for half the problem - 'collapse' (content from high to low height) will animate but 'expand' (low to high) will not

then I have a very hacky solution that sets the height to -2 then use a bottomSheetCallback monitoring the first time onSlide triggers with offset = 0 (finished sliding) to set layout height. it works both ways now, but its quite a mess.


solution 1: (see how onSlide does not log on expand anymore)


[Screencast from 07-08-2025 04:22:24 PM.webm](https://github.com/user-attachments/assets/77fba042-60ff-40e0-95a6-04e28f459db8)


solution 2:


[Screencast from 07-08-2025 04:21:30 PM.webm](https://github.com/user-attachments/assets/c194711e-f81e-4614-b08b-7605e03fc7a8)

lodev09:
Have you tried intercepting touch gesture?

lovegaoshi:
im not comfortable writing an ontouch intercept logic to effectively disable bottomsheet.behavior.EXPANDED, ideas welcome

i did set maxHeight to -1 instead of value + footerHeight in the 2nd approach - if view.layout.height just has to be slightly higher than peekHeight to trigger the animation, this could be an easy solution. otherwise i guess fixing the layout height in onSlide is the only way out

lovegaoshi:
i spent a bit more time and found some leads. though im sorta lost here and need another pair of eyes.

I did confirm my hypothesis that view.layout.height just has to be slightly higher than peekHeight to trigger the animation - presented in solution 1, I override truesheetview.layout.height to peekHeight + 1 to give an animation. but this animation is janky in my testings. expand doesnt expand quite to the exact hight and collapse is too fast.

the solution 2 is the hacky solution i mentioned before - only override truesheetview.layout.height to peekheight AFTER the first completed onSlide (which is the animation trigger from content size change). this way bc collapse no longer overrides the layout up front, the animation is a lot smoother; and for unexplained reasons now the expand can also expand to full length. the downside is extra complicated code logic and obviously user scrolls during animation will have side effects. of course I can also do a isDraggable = false then set it to true.

lodev09:
@lovegaoshi I agree, animating size change is too hacky. Let's just stick with no-animation approach. I've committed to simplify things, hope you don't mind :)

lovegaoshi:
alright. I'll probably adapt this code in my fork instead. I like
animations:)

lodev09:
Go for it! Although you might as well use `['auto', 'large']` if you really want that animation.
I did try various combinations but couldn't find any stable solution. Test it using the `PrompSheet` example :)

lodev09:
:rocket: _This pull request is included in v2.0.6. See [Release 2.0.6](https://github.com/lodev09/react-native-true-sheet/releases/tag/v2.0.6) for release notes._

PR #199, opened 2025-07-09
feat: add support for headers and change the layout approach

Cyberax:
This commit adds support for header components. However, the major improvement is the changed approach to doing the layout for the bottom sheet.

Prior to this commit, there was an inherent conflict between the native iOS/Android layout systems and the React layout manager. In particular, this resulted in issues with the on-screen keyboard overlapping controls.

To solve this, this commit introduces an intermediate view that sits between the native OS dialog and the hosted React view. This intermediate view handles the keyboard avoidance using OS-specific mechanism: keyboardLayoutGuide on iOS and rootView.padding on Android.

We then communicate the visible area back to React that then lays out components as usual. So there is no need to have special handling for scrolling views or to do manual footer positioning.


vilindberg:
@Cyberax Thanks for this! Also having issues with both keyboard handling and footer positioning. 

Any more work planned on this PR? Otherwise I'll give this a go and see if it solved my issues. In that case, are there any obvious TODO's I should look out for?

GaeCuure:
Any update on this ?

lodev09:
Just released #211 beta. Please test it out
closing this for now

PR #216, opened 2025-11-24
fix: android build error `Too many arguments for...`

camchis:
Android builds are failing with error

```
Too many arguments for 'fun onChildStartedNativeGesture(androidEvent: MotionEvent, eventDispatcher: EventDispatcher): Unit'.
```
This PR removes the unnecessary param being passed into `onChildStartedNativeGesture`


lodev09:
What version of android?

camchis:
> What version of android?
```
    compileSdkVersion = 36
    targetSdkVersion = 35
```
@lodev09

lodev09:
Sorry, I meant RN version

camchis:
@lodev09 0.79.7

camchis:
ah, I can see min supported version is 0.81! will stay on v2 for now then

lodev09:
If you already in new arch, use v3! you can just patch it :)

lodev09:
can confirm this was the right fix. created separate PR since I have to remove the annotations as well
#231

PR #226, opened 2025-11-25
feat(android): use Material Design 3 defaults

lodev09:
## Summary

Improves Android's default styling to align with Material Design 3 guidelines and iOS behavior.

Addresses #222

### Corner Radius
- Changed default `cornerRadius` from `0` to `28dp` (Material Design 3 bottom sheet spec)
- iOS already uses the device's native corner radius when not provided

### Background Color
- Changed default `backgroundColor` from `WHITE` to `colorSurfaceContainerLow`
- Automatically adapts to light/dark mode
- Falls back to `WHITE` if theme attribute is unavailable

### Documentation
- Updated props reference to explain platform-specific defaults for both `cornerRadius` and `backgroundColor`

### Example Code
- Removed `cornerRadius` from README and example since defaults now work out of the box

## Why

Previously, users had to use platform-specific code like:
```tsx
cornerRadius={Platform.OS === 'android' ? 30 : undefined}

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
@@ -196,7 +196,7 @@ class TrueSheetViewController(private val reactContext: ThemedReactContext) :
     }
 
   var cornerRadius: Float = 28f.dpToPx()
-  var sheetBackgroundColor: Int = getDefaultBackgroundColor()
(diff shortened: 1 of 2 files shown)


lodev09:
:rocket: _This pull request is included in v3.0.0-beta.6. See [Release 3.0.0-beta.6](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.0.0-beta.6) for release notes._

PR #256, opened 2025-11-29
fix: improve detent precision, interpolation, and native safe area handling

lodev09:
## Summary

- Use `Double` instead of `Float` for detent values to preserve precision (fixes long floating point values like `0.15000000596046448`)
- Add `interpolatedDetentForPosition` method on iOS and Android
- Rename `detent` to `animatedDetent` in Reanimated provider
- `animatedDetent` now animates/interpolates smoothly like `animatedIndex`
- Remove `insetAdjustment` workaround and let iOS handle bottom insets natively

### iOS
- Changed `floatValue` to `doubleValue` when reading detent values
- Added `interpolatedDetentForPosition:` method for smooth detent interpolation
- Removed `insetAdjustment` - iOS now handles bottom safe area insets natively
- Position is now the source of truth for calculating interpolated index and detent values

### Android
- Added `getInterpolatedDetentForPosition` method

### Codegen
- Changed `detents` array type from `Float` to `Double`

### Reanimated
- Renamed `detent` → `animatedDetent` with smooth animation support

## Breaking Change: Bottom Safe Area Handling

Previously, the library internally adjusted for bottom safe area insets on iPhone. This created inconsistencies across devices (iPhone vs iPad/visionOS) and made the position calculations complex.

**What changed:**
- iOS now handles bottom safe area insets natively (no manual adjustment)
- The `position`, `index`, and `detent` values are now calculated purely from the actual sheet position relative to screen height
- This means on iPhone, a sheet at detent `0.1` will actually be taller than `0.1 * screenHeight` because iOS adds the bottom safe area internally

**Migration for footer with auto detent:**
If you have a sheet with `footer` prop and `auto` detent, you may need to account for the bottom safe area in your footer component:

```tsx
import { useSafeAreaInsets } from 'react-native-safe-area-context';

const Footer = () => {
  const insets = useSafeAreaInsets();
  
  return (
    <View style={{ paddingBottom: insets.bottom, backgroundColor: '#333' }}>
      <View style={{ height: 60, justifyContent: 'center', alignItems: 'center' }}>
        <Text>Footer Content</Text>
      </View>
    </View>
  );
};
```

This ensures the footer extends into the safe area while keeping content above it.

### DIFF
--- a/example/src/screens/MapScreen.tsx
+++ b/example/src/screens/MapScreen.tsx
@@ -21,7 +21,6 @@ import Animated, { useAnimatedStyle } from 'react-native-reanimated';
 import { Button, Header, Spacer } from '../components';
 import { BLUE, DARK, GAP, GRAY, HEADER_HEIGHT, SPACING } from '../utils';
 
-import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import {
   BasicSheet,
   BlankSheet,
@@ -35,7 +34,6 @@ import { useAppNavigation } from '../hooks';
 const AnimatedButton = Animated.createAnimatedComponent(TouchableOpacity);
 
 export const MapScreen = () => {
-  const insets = useSafeAreaInsets();
   const { height } = useWindowDimensions();
   const { animatedPosition } = useReanimatedTrueSheet();
   const navigation = useAppNavigation();


lodev09:
:rocket: _This pull request is included in v3.0.0-beta.18. See [Release 3.0.0-beta.18](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.0.0-beta.18) for release notes._

Issue #255, opened 2025-11-29
animatedIndex issues on iOS


RazvanGorea:
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

In the following examples the expected behavior is for the red box to have opacity 0 on first detent and 1 on the second. On the iPhone 17 pro animatedIndex only breaks for sheets smaller than approx. detent of 0.21. On the iPad it is broken regardless of the sheet height. I'm almost sure both cases have something to do with bottom inset.

https://github.com/user-attachments/assets/592f3fb1-8acd-436f-b350-319fd52e1d9a

https://github.com/user-attachments/assets/68311e73-f853-490b-94b5-6fc6ae7a9186

Here's the code from the examples:

```
import {
  ReanimatedTrueSheet,
  TrueSheet,
  useReanimatedTrueSheet,
} from "@lodev09/react-native-true-sheet";
import { useRef } from "react";
import { Button, StyleSheet, View } from "react-native";
import Animated from "react-native-reanimated";

export default function Screen() {
  const sheetRef = useRef<TrueSheet>(null);

  const { animatedIndex } = useReanimatedTrueSheet();

  return (
    <View style={styles.container}>
      <Button
        title="Open Sheet"
        onPress={() => {
          sheetRef.current?.present();
        }}
      />

      <ReanimatedTrueSheet ref={sheetRef} dimmed={false} detents={[0.1, 0.15]}>
        <Animated.View
          style={[
            { height: 100, backgroundColor: "red", opacity: animatedIndex },
          ]}
        />
      </ReanimatedTrueSheet>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
});
```

### Library version

3.0.0-beta.17

### Environment info

```shell
N\A
```

### Steps to reproduce

1. Check the repo below

### Reproducible example repository

https://github.com/RazvanGorea/true-sheet-animation-repro

lodev09:
This is the log I'm getting when I disabled Liquid Glass:
```
eight: 874, position: 742.99814212573, animatedIndex: 0.9977118349191714, detent: 0.15000000596046448
height: 874, position: 742.9985962478929, animatedIndex: 0.9977117656368252, detent: 0.15000000596046448
[...]

PR #262, opened 2025-12-01
fix(ios): emit -1 position when sheet is going to dismiss

lodev09:
Followup fix for #255


lodev09:
:rocket: _This pull request is included in v3.0.1. See [Release 3.0.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.0.1) for release notes._

PR #265, opened 2025-12-01
fix(ios): skip presenting VC that is being dismissed

lodev09:
## Summary

When presenting a sheet on screen focus (e.g., using `useFocusEffect`) while a modal is being dismissed, the sheet could present on the wrong view controller - the one being dismissed.

## Changes

- Updated `findPresentingViewController` in `TrueSheetView.mm` to check `isBeingDismissed` flag and skip those view controllers
- Added documentation for the `requestAnimationFrame` workaround when presenting on screen focus
- Updated example to demonstrate the workaround

## Note

The `isBeingDismissed` flag is only set after the dismissal animation begins. When using React Navigation's `useFocusEffect`, the focus event fires before the animation starts. Using `requestAnimationFrame` gives iOS enough time to mark the view controller as being dismissed.

### DIFF
--- a/example/src/screens/MapScreen.tsx
+++ b/example/src/screens/MapScreen.tsx
@@ -1,4 +1,4 @@
-import { useRef, useState, useCallback } from 'react';
+import { useRef, useState } from 'react';
 import {
   Platform,
   StyleSheet,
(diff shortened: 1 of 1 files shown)


lodev09:
:rocket: _This pull request is included in v3.0.3. See [Release 3.0.3](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.0.3) for release notes._

PR #268, opened 2025-12-02
Fix android back button press when dismissible false

jymaa:
Add back button handling to dialog.

To test: 

- Add a Sheet with `dismissible={false}`
- Try to use back button on Android
- Back button should work

Not a android dev expert so, you might want to change some things. But i've tried it and it work for me.


lodev09:
Closing in favor of #269

PR #296, opened 2025-12-09
fix(grabber): use ProcessedColorValue for color prop and fix z-order

lodev09:
### Color type fix
- Changed `grabberOptions.color` from `Int32` to `ProcessedColorValue | null` in the native component spec
- Added `processColor()` in TrueSheet.tsx to convert colors before passing to native
- Updated iOS native code to use `RCTUIColorFromSharedColor` directly (removed `SharedColor()` wrapper)
- Updated Android null check to use `isNull()` instead of comparing to `0`

### Z-order fix (iOS only)
- Added `bringSubviewToFront` for custom grabber view to ensure it renders above header views

Android z-order is already correct since the grabber is added after the content view.

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
@@ -130,7 +130,7 @@ class TrueSheetViewManager :
       } else {
         null
       },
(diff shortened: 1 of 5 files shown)


lodev09:
:rocket: _This pull request is included in v3.2.2. See [Release 3.2.2](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.2.2) for release notes._

PR #297, opened 2025-12-09
fix: interpolated index and detent for single detent

lodev09:
## Problem
When only one detent is provided (e.g., `[1]`), the interpolated `index` and `detent` values emitted in position change events were fixed at `0` and `1` respectively, instead of changing as the sheet moves.

## Solution
Updated `findSegmentForPosition` on both iOS and Android to calculate the interpolation between the closed state and the single detent.

Also refactored to deduplicate the 'above first detent' logic by handling it once before the single/multi-detent branching.

### DIFF
--- a/example/src/screens/MapScreen.tsx
+++ b/example/src/screens/MapScreen.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { useRef, useState } from 'react';
 import {
   Platform,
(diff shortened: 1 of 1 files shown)


lodev09:
:rocket: _This pull request is included in v3.2.2. See [Release 3.2.2](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.2.2) for release notes._

PR #323, opened 2025-12-14
fix: key window fallback for cold start and deep link handling

### LINKED ISSUE #321: Sheet wont open when entering the app from a deep app link

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

I have this _layout in my app:
```
import { createTrueSheetNavigator } from "@lodev09/react-native-true-sheet/navigation";
import { withLayoutContext } from "expo-router";

const { Navigator } = createTrueSheetNavigator();
export default withLayoutContext(Navigator);

```

and a route under that _layout named favorites

when opening the app from a widget link of: <app-name>://favorites, i see the app opening on the right route (using the pathnames) altough the sheet in this route wont open

in the route i have set the options as follow:
```
useEffect(() => {
    navigation.setOptions({
      detents: [0.5],
      insetAdjustment: "never",
      detentIndex: 0,
      reanimated: true,
      scrollable: true,
      dismissible: true,
    });
  }, [navigation]);
```

am i doing something wrong?

### Library version

3.3.1

### Environment info

```shell
System:
  OS: macOS 26.0.1
  CPU: (11) arm64 Apple M3 Pro
  Memory: 112.02 MB / 18.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.15.1
    path: /Users/eagle/.nvm/versions/node/v20.15.1/bin/node
  Yarn:
    version: 1.22.22
    path: /Users/eagle/.bun/bin/yarn
  npm:
    version: 10.7.0
    path: /Users/eagle/.nvm/versions/node/v20.15.1/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.16.2
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.0
      - iOS 26.0
      - macOS 26.0
      - tvOS 26.0
      - visionOS 26.0
      - watchOS 26.0
  Android SDK: Not Found
IDEs:
  Android Studio: Not Found
  Xcode:
    version: 26.0.1/17A400
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 25.0.1
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
    installed: 20.0.2
    wanted: ^20.0.2
  react:
    installed: 19.1.0
    wanted: 19.1.0
  react-native:
    installed: 0.81.5
    wanted: 0.81.5
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

### Steps to reproduce

1. create the layout as described
2.at a route with the following options under that layout
3. open the app using an app link to that route

### Reproducible example repository

-

lodev09:
## Summary

Fixes #321 - Sheet won't open when entering the app from a deep link

## Problem

When launching the app via deep link (e.g., `<app-name>://favorites`), the app navigates to the correct route but the sheet fails to display. This happens because:

1. **iOS**: During cold start or deep link handling from background, no scene is in `UISceneActivationStateForegroundActive` state yet, so `keyWindow` returns `nil` and no presenting view controller is found
2. **Navigation**: Deep links bypass the base route, breaking the expected route structure that `TrueSheetView` relies on (`const [baseRoute, ...sheetRoutes] = state.routes`)

### iOS (`WindowUtil.mm`)
- Added fallback window lookup for transitional states
- Still prefers foreground active scenes, but falls back to any available key window
- Returns a valid window instead of `nil` during app launch transitions

### Navigation (`TrueSheetRouter.ts`)
- Ensures base route is always present in `getInitialState` and `getRehydratedState`
- Extracted common logic into `ensureBaseRoute` helper
- Inserts base route at index 0 when missing (deep link scenarios)
- Correctly sets index to point to the deep link target

## Testing

- Verified both call sites of `keyWindow` already handle `nil` gracefully
- Router changes align with `TrueSheetView`'s expectation: `const [baseRoute, ...sheetRoutes] = state.routes`
- goBack/dismiss logic properly protects base route at index 0

### DIFF
--- a/ios/utils/WindowUtil.mm
+++ b/ios/utils/WindowUtil.mm
@@ -15,10 +15,13 @@
 + (nullable UIWindow *)keyWindow {
   NSSet<UIScene *> *connectedScenes = [UIApplication sharedApplication].connectedScenes;
 
+  UIWindow *fallbackWindow = nil;
+
   for (UIScene *scene in connectedScenes) {
     if ([scene isKindOfClass:[UIWindowScene class]]) {
       UIWindowScene *windowScene = (UIWindowScene *)scene;
 
+      // Check for foreground active first (preferred)
       if (windowScene.activationState == UISceneActivationStateForegroundActive) {
         for (UIWindow *window in windowScene.windows) {
           if (window.isKeyWindow) {
@@ -31,10 +34,23 @@
           return windowScene.windows.firstObject;
         }
       }
+
+      // Store a fallback window for transitional states (e.g., deep link handling from background or cold start)
+      if (fallbackWindow == nil && windowScene.windows.count > 0) {
+        for (UIWindow *window in windowScene.windows) {
+          if (window.isKeyWindow) {
+            fallbackWindow = window;
+            break;
+          }
+        }
+        if (fallbackWindow == nil) {
+          fallbackWindow = windowScene.windows.firstObject;
+        }
+      }
     }
   }
 
-  return nil;
+  return fallbackWindow;
 }
 
 @end
--- a/src/navigation/TrueSheetRouter.ts
+++ b/src/navigation/TrueSheetRouter.ts
@@ -40,6 +40,34 @@ export const TrueSheetActions = {
   remove: (): TrueSheetActionType => ({ type: 'REMOVE' }),
 };
 
+const ensureBaseRoute = <T extends { routes: { name: string }[] }>(
+  state: T,
+  baseRouteName: string | undefined,
+  routeParamList: Record<string, object | undefined> | undefined
+): T & { index: number; routes: T['routes'] } => {
+  if (!baseRouteName) {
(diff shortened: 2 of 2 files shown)


lodev09:
:rocket: _This pull request is included in v3.3.3. See [Release 3.3.3](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.3.3) for release notes._

Bram-dc:
Could it be that this is broken again? I am having the same issue on the newer version.

PR #365, opened 2025-12-29
fix(android): only handle keyboard for focused views within sheet

### LINKED ISSUE #364: Android IME pushes TrueSheet even when focused TextInput is outside the sheet (v3.6.8)
### Discussed in https://github.com/lodev09/react-native-true-sheet/discussions/362

<div type='discussions-op-text'>

<sup>Originally posted by **rovieradev** December 29, 2025</sup>
Hi!
I’m currently using react-native-true-sheet v3.6.8 and I’m running into an Android-specific keyboard (IME) issue.

Important detail: the focused TextInput is not inside the TrueSheet.

When the keyboard appears, the entire TrueSheet is pushed upward, even though:

- the sheet size does not change
- the input lives outside the sheet hierarchy
- detents and layout are not being updated manually

It feels like the system is applying IME insets (adjustResize) at the window level, and since TrueSheet is attached to android.R.id.content, the sheet moves even though it’s unrelated to the focused input.

I’ve tried:

- different keyboardMode values (pan / resize)
- changing windowSoftInputMode
- overlay-style handling on the app side

…but the behavior persists.

One thing worth noting: closing the keyboard using the Android system “hide keyboard” button reproduces the issue consistently, which makes it seem like this happens at the window / IME inset level, before TrueSheet logic runs.

Expected behavior:
The sheet should remain anchored and only react to keyboard changes when the focused input is actually inside the sheet (similar to iOS).

Is this a known limitation of the current Android implementation, or is there a recommended setup to prevent the sheet from being pushed by IME when the input is external?

Thanks!</div>

lodev09:
Fixes #364

This PR ensures the sheet only responds to keyboard events when the focused view is within the sheet itself. Previously, focusing an input outside the sheet (e.g., in an underlying screen) would still cause the sheet to adjust.

## Changes

- Added `isFocusedViewWithinSheet()` method to `TrueSheetKeyboardObserver` that checks if the currently focused view is a descendant of the sheet
- Updated `shouldHandleKeyboard()` to check focus for `keyboardWillShow` but skip the check for `keyboardWillHide` (since focus is already cleared when keyboard dismisses)
- Moved `isKeyboardTransitioning` state to the observer for better encapsulation

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
@@ -240,6 +240,11 @@ class TrueSheetViewController(private val reactContext: ThemedReactContext) :
   private val isKeyboardTransitioning: Boolean
     get() = keyboardObserver?.isTransitioning ?: false
 
+  private fun isFocusedViewWithinSheet(): Boolean {
+    val sheet = sheetView ?: return false
+    return keyboardObserver?.isFocusedViewWithinSheet(sheet) ?: false
+  }
+
   val bottomInset: Int
     get() = if (edgeToEdgeEnabled) ScreenUtils.getInsets(reactContext).bottom else 0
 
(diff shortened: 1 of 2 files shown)


lodev09:
:rocket: _This pull request is included in v3.6.9. See [Release 3.6.9](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.6.9) for release notes._

PR #369, opened 2025-12-30
fix(android): allow sheet drag when ScrollView cannot scroll

lodev09:
## Summary

When a `ScrollView` inside the sheet has content smaller than the viewport, it cannot scroll and previously blocked sheet dragging. This PR adds touch interception to allow the sheet to be dragged in these cases.

## Changes

- Added touch interception in `TrueSheetCoordinatorLayout` to detect non-scrollable `ScrollView`s and allow sheet dragging
- Only enabled when `scrollable` prop is `true`
- Removed the workaround documentation from troubleshooting docs

## Inspiration

This is a workaround until [NestedScrollView is merged into react-native core](https://github.com/facebook/react-native/pull/44099).


lodev09:
:rocket: _This pull request is included in v3.6.10. See [Release 3.6.10](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.6.10) for release notes._

PR #372, opened 2025-12-31
fix(android): propagate eventDispatcher via delegate chain for footer touch handling

lodev09:
## Summary

Fixes an issue where footer touch events don't work when the footer is added at runtime via `navigation.setOptions`.

## Problem

When using `navigation.setOptions({ footer: ... })` to add a footer after the sheet is already presented, the footer's `eventDispatcher` was never set, causing touch events to not be dispatched to React Native.

## Solution

Refactored the `eventDispatcher` to flow through the delegate chain:

- `TrueSheetView` (host) owns `eventDispatcher` (set by ViewManager)
- `TrueSheetViewController` accesses it via `delegate?.eventDispatcher`
- `TrueSheetContainerView` accesses it via `delegate?.eventDispatcher`
- `TrueSheetFooterView` accesses it via `delegate?.eventDispatcher`

This ensures the footer always has access to the current `eventDispatcher` regardless of when it's added.

## Test

Added a test case in `DetailsSheet` that sets a footer via `useEffect` + `navigation.setOptions`. Tapping the footer presents a nested sheet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ The true native bottom sheet experience for your React Native Apps. 💩
 
 - React Native >= 0.76 (Expo SDK 52+)
 - New Architecture enabled (default in RN 0.76+)
+- Xcode 26.2 (strongly recommended for better library functionality)
 
 ### Expo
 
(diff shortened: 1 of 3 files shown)


lodev09:
:rocket: _This pull request is included in v3.6.11. See [Release 3.6.11](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.6.11) for release notes._

PR #405, opened 2026-01-10
fix(web): fire onWillDismiss when dismissing via backdrop tap

### LINKED ISSUE #401: [web] OnWillDismiss isn't called

### Bug Summary

Hi, first of all thanks for the amazing work. I found that on web platform `onWillDismiss` isn't called when user clicked outside the sheet. To be clear on web `onWillDismiss` works only if you close a sheet by `.dismiss()` call. On Android it works in both cases.

p.s. `onDidDismiss` works fine in both cases in web/android.

### Affected Platforms

- [ ] iOS
- [ ] Android
- [x] Web
- [ ] Other

### Library Version

3.7.0

### Environment Info

```shell
react-native 0.82.1
```

### Steps to Reproduce

1. `onWillDismiss={() => console.log('onWillDismiss')}`
2. open the sheet
3. click outside the sheet to dismiss it
4. look at console logs

### Repro

-

### Additional Context

_No response_

lodev09:
## Summary
Fix `onWillDismiss` not being called when the sheet is dismissed by clicking outside (backdrop tap) on web.

The callback was only firing during programmatic `.dismiss()` calls because it was gated behind the `isDismissing` flag. Now it fires whenever the sheet animates to index `-1`, regardless of how the dismiss was triggered.

Fixes #401

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Test Plan
1. Open a sheet on web
2. Click outside the sheet to dismiss
3. Verify `onWillDismiss` is called

## Checklist
- [x] Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### 🐛 Bug fixes
 
+- **Web**: Fixed `onWillDismiss` not firing when dismissing via backdrop tap. ([#405](https://github.com/lodev09/react-native-true-sheet/pull/405) by [@lodev09](https://github.com/lodev09))
 - Fixed sheet not dismissing when presenter screen is popped from navigation. ([#400](https://github.com/lodev09/react-native-true-sheet/pull/400), [#402](https://github.com/lodev09/react-native-true-sheet/pull/402) by [@lodev09](https://github.com/lodev09))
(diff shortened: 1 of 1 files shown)


lodev09:
:rocket: _This pull request is included in v3.7.1. See [Release 3.7.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.7.1) for release notes._

PR #406, opened 2026-01-11
fix(ios): verify window during initial present

lodev09:
## Summary

Fixes an issue where the sheet would fail to present when navigating via deeplinks. The view controller's window may not match the sheet's window during navigation transitions.

### Changes:
- Verify the view controller is in the same window before initial present
- Check that the view controller is not being dismissed
- Reset `_didInitiallyPresent` flag when view is removed from superview
- Log warning when initial present is skipped

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test Plan

1. Open app via deeplink that navigates to a screen with a sheet using `initialDetentIndex`
2. Sheet should present correctly after navigation completes

## Checklist

- [x] I have tested this on iOS
- [ ] I have tested this on Android (N/A - iOS only change)
- [ ] I have tested this on Web (N/A)
- [ ] I have updated the documentation (if applicable)
- [ ] I have updated the changelog (will update after PR created)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed initial present failing during deeplink navigation. ([#406](https://github.com/lodev09/react-native-true-sheet/pull/406) by [@lodev09](https://github.com/lodev09))
+
 ## 3.7.2
 
 ### 🎉 New features


lodev09:
:rocket: _This pull request is included in v3.7.3. See [Release 3.7.3](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.7.3) for release notes._

PR #416, opened 2026-01-13
fix(ios): improve willDismiss event timing during drag dismiss

### LINKED ISSUE #411: [iOS] `onWillDismiss` fires prematurely during interactive drag gesture

### Bug Summary

When a user interactively drags a sheet to dismiss it, `onWillDismiss` fires as soon as the drag crosses the dismiss threshold—even if the user then drags the sheet back up and cancels the dismiss.

The callback fires at the **start** of the potential dismiss, not when the dismiss is actually **confirmed** (i.e., when the user releases their finger).

### Affected Platforms

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Other

### Library Version

3.7.2

### Environment Info

```shell
## Environment

- react-native-true-sheet version: 3.7.2
- React Native version: 0.81.5
- Platform: iOS 
- Expo SDK: 54
```

### Steps to Reproduce

1.  Drag sheet down past the dismiss threshold
2. `onWillDismiss` fires immediately when threshold is crossed
3. Drag sheet back up (without releasing)
4. Releases finger - sheet stays open
5. `onWillDismiss` was fired but dismiss never happened

### Repro

https://github.com/szt217/true-sheet-issue-demo

## Expected Behavior

`onWillDismiss` should only fire when the dismiss is **confirmed** - meaning:
- The user has released their finger AND
- The sheet will actually proceed to dismiss

This would make it a reliable signal that the sheet **will** dismiss, not that it **might** dismiss.

## Use Case

I need to perform cleanup/preparation actions when the sheet is about to close. With the current behavior, I can't reliably use `onWillDismiss` because:

1. If I trigger side effects in `onWillDismiss`, they happen even when the user cancels the dismiss
2. I have no way to know if the dismiss was cancelled after `onWillDismiss` fired
3. I can't use `onDismiss` (onDidDismiss) because that fires too late - after the animation completes

## Proposed Solution

The ideal behavior would be:

- **`onWillDismiss`**: Fires when the user **releases their finger** and the sheet **will** proceed to dismiss (but animation hasn't started yet)
- **`onDismiss`**: Fires when the dismiss animation **completes** (current behavior)

Alternatively, a new callback like `onDismissGestureStart` could fire when the threshold is crossed, while `onWillDismiss` fires on confirmed dismiss.

## iOS Implementation Context

On iOS, `UISheetPresentationController` provides the `sheetPresentationControllerWillDismiss` delegate method. The timing of when this fires during interactive gestures may be the root cause. You may need to track the gesture state to only fire the JS callback when the gesture ends with a confirmed dismiss.

lodev09:
## Summary
Fixes willDismiss event timing during drag-to-dismiss gesture. The event now fires when the sheet position reaches the screen height during drag, rather than only in viewWillDisappear.

## Type of Change
- [x] Bug fix

## Test Plan
1. Open a sheet
2. Drag to dismiss
3. Verify willDismiss fires at the correct time (when sheet reaches bottom)

Fixes #411

## Checklist
- [x] iOS
- [ ] Android (N/A)
- [ ] Web (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## Unreleased
 
-### 🎉 Bug fixes
+### 🐛 Bug fixes
 
 - Auto re-present sheet when returning from screen dismiss (modal or navigation pop). ([#412](https://github.com/lodev09/react-native-true-sheet/pull/412) by [@lodev09](https://github.com/lodev09))
+- **iOS**: Fixed `onWillDismiss` event timing during drag dismiss. ([#416](https://github.com/lodev09/react-native-true-sheet/pull/416) by [@lodev09](https://github.com/lodev09))
 
 ### 💡 Others
 


lodev09:
:rocket: _This pull request is included in v3.7.4-beta.0. See [Release 3.7.4-beta.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.7.4-beta.0) for release notes._

lodev09:
:rocket: _This pull request is included in v3.8.0. See [Release 3.8.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.8.0) for release notes._

PR #451, opened 2026-01-19
fix: use measured footer height for keyboard scroll offset

Layrton:
# fix: use measured footer height for keyboard scroll offset

## Summary

When using a sheet with a footer and `keyboardScrollOffset` is not explicitly set (or set to 0), inputs near the bottom of the ScrollView get hidden behind the footer when the keyboard appears.

This PR makes the native layer measure the actual footer height and use it automatically for keyboard scroll calculations.

### Problem

The current implementation relies on JS-provided `keyboardScrollOffset` to account for footer height when scrolling inputs into view. However:

1. Users often don't know they need to set this prop
2. Calculating the correct value requires knowing the exact rendered footer height (which may differ from layout constants due to safe area insets, dynamic content, etc.)
3. The example app's `PromptSheet` was passing an incorrect value (`FOOTER_HEIGHT + SPACING = 64`) while the actual rendered footer was ~82pt

### Solution

The native code now measures the actual footer height and uses it when `keyboardScrollOffset` is 0 or not provided:

```
scrollOffset = keyboardScrollOffset > 0 ? keyboardScrollOffset : footerHeight
```

**Behavior**:

- If JS provides `keyboardScrollOffset > 0` → use it (fully backwards compatible)
- Otherwise → use measured `footerHeight` automatically

This makes footers "just work" without requiring manual measurement or prop configuration.

## Type of Change

- [x] Bug fix

## Test Plan

Tested with the example app's PromptSheet:

1. Open sheet at any detent (0.5, 0.85, 1.0)
2. Tap input field near bottom of the ScrollView
3. Verify input scrolls into view above footer when keyboard appears
4. Verify backwards compatibility: if `keyboardScrollOffset` is explicitly set, it takes precedence

**Platforms tested:**

- iOS 26 (iPhone 17 Pro Max simulator)
- Android (Pixel emulator)

### Before (input hidden behind footer)

https://github.com/user-attachments/assets/6bd7de44-926a-49e0-9560-3be0a350b2ca

### After (input scrolls above footer)

https://github.com/user-attachments/assets/79c267a8-9314-4c28-90c7-d2ade0f5274c

## Checklist

- [x] I have tested this on **iOS**
- [x] I have tested this on **Android**
- [ ] I have tested this on **Web** (N/A - native-only change)
- [ ] I have updated the documentation accordingly (N/A - no API changes)
- [x] I have added an entry to `CHANGELOG.md`

## Changes

**iOS:**

- `TrueSheetContentView.h` - Added `footerHeight` property
- `TrueSheetContentView.mm` - Uses footer height in scroll calculation
- `TrueSheetContainerView.mm` - Passes footer height to content view during layout

**Android:**

- `TrueSheetContentView.kt` - Added `footerHeight` property and scroll logic
- `TrueSheetContainerView.kt` - Passes footer height to content view

**Example:**

- `PromptSheet.tsx` - Removed manual `keyboardScrollOffset` (no longer needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### 🐛 Bug fixes
 
+- **iOS, Android**: Fixed keyboard scroll offset to use measured footer height when `keyboardScrollOffset` is not provided. ([#451](https://github.com/lodev09/react-native-true-sheet/pull/451) by [@Layrton](https://github.com/Layrton))
 - **Android**: Fixed present promise not resolving on resize. ([c3495500](https://github.com/lodev09/react-native-true-sheet/commit/c3495500) by [@lodev09](https://github.com/lodev09))
 - **iOS**: Fixed scroll view pinning to respect content view padding/margin. ([#429](https://github.com/lodev09/react-native-true-sheet/pull/429), [#446](https://github.com/lodev09/react-native-true-sheet/pull/446) by [@lodev09](https://github.com/lodev09))
 - **iOS**: Fixed footer not translating back when keyboard hides via ScrollView. ([#424](https://github.com/lodev09/react-native-true-sheet/pull/424) by [@lodev09](https://github.com/lodev09))


lodev09 (commented):
Hmmm, I'm not sure about this. Some users will use footer as a floating component like a FAB. Adding this as default will be an issue.

lodev09:
Closing this. measuring footer height in JS and using it as offset shouldn't be that hard.

PR #444, opened 2026-01-18
feat: add pop to cascade dismiss behavior

obi-owner:
## Summary

Fixes #443 

#452 must be merged before this PR can be merged

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Created examples for expo, bare, and web. Tested that existing dismiss behavior still works and new popTo and popToTop behavior works

## Screenshots / Videos

https://github.com/user-attachments/assets/d04d4252-7431-479c-ac8d-b98dadfa2078

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [x] I tested on Web
- [x] I updated the documentation (if needed)
- [ ] I added a changelog entry (if needed)

PR #452, opened 2026-01-20
feat: add native cascade dismiss api

obi-owner:
## Summary

- Refactors the native dismiss API to dismiss the current sheet and all child sheets presented on top of it when `dismiss()` is called
- Adds a `dismissChildren()` method to only dismiss the children presented on top of the current sheet

This must be merged before #444 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [x] I tested on Web

### DIFF
--- a/docs/docs/guides/onmount.mdx
+++ b/docs/docs/guides/onmount.mdx
@@ -32,6 +32,8 @@ const App = () => {
 
 You may want to disable the present animation. To do this, simply set [`initialDetentAnimated`](/reference/configuration#initialdetentanimated) to `false`.
 
-### Using with React Navigation
+### Deep Linking on iOS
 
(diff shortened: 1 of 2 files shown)


lodev09 (changes requested):
Looks good so far. Just need to retain the `dismiss` name on the native instead of the old `dismissAll`. This should make methods consistent in the native too. Just add code comment so it's clear.

lodev09:
Also, I think `dismissStack` is more proper, let's go with that 😄

lodev09:
:rocket: _This pull request is included in v3.8.0. See [Release 3.8.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.8.0) for release notes._

PR #478, opened 2026-01-27
fix(web): correct event firing order and conditions

lodev09:
## Summary
Fix web event firing to match native behavior:

- `onWillDismiss` no longer fires when a new sheet is presented on top (only fires on actual dismiss)
- `onMount` now fires on first present instead of component mount
- Event order now matches native: mount → willPresent → willFocus → didPresent → didFocus

## Type of Change
- [x] Bug fix

## Test Plan
Tested on web example app with stacked sheets and verified event order in console.

## Checklist
- [ ] iOS
- [ ] Android
- [x] Web


lodev09:
:rocket: _This pull request is included in v3.8.1. See [Release 3.8.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.8.1) for release notes._

PR #479, opened 2026-01-27
fix(web): collapse to lowest detent when not dismissible

lodev09:
## Summary
When `dismissible=false`, backdrop tap now collapses to `dimmedDetentIndex - 1` instead of doing nothing, matching iOS/Android behavior.

## Type of Change
- [x] Bug fix

## Test Plan
Tested on web with `dismissible={false}` and multiple detents.

## Checklist
- [x] Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 📖 Documentation
+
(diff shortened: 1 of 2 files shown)


lodev09:
:rocket: _This pull request is included in v3.8.1. See [Release 3.8.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.8.1) for release notes._

PR #482, opened 2026-01-28
fix(ios): clear stale ScrollView reference when content changes

### LINKED ISSUE #481: Stale ScrollView reference causes content clipping when sheet content changes
## Description

When the content of a TrueSheet changes dynamically (e.g., switching between different content components), the sheet may hold a stale reference to a ScrollView that is no longer in the view hierarchy. This causes content to be clipped/cut off until a user interaction (like dragging the sheet) triggers a re-layout.

## Steps to Reproduce

1. Create a TrueSheet with dynamic content that switches between different views (e.g., SearchContent → SettingsContent)
2. Both content views contain ScrollViews
3. Switch from one content to another
4. Observe that the new content is clipped at the bottom
5. Drag the sheet grabber and release - content becomes fully visible

## Expected Behavior

Content should be fully visible immediately when sheet content changes.

## Actual Behavior

Content is clipped on initial render. Only after a user gesture (drag and release) does the content become fully visible.

## Root Cause

In `TrueSheetContentView.mm`, the `setupScrollable:bottomInset:` method has an early return optimization at lines 102-105:

```objc
// Already set up with same inset
if (_pinnedScrollView && _bottomInset == bottomInset) {
  return;
}
```

This check doesn't verify that `_pinnedScrollView` is still valid (still in the view hierarchy). When content changes, the old ScrollView is unmounted but the reference is retained, causing `updateScrollViewHeight` to operate on a stale/invalid view.

## Proposed Fix

Add a validity check before the early return:

```objc
// Check if pinned scroll view is still valid (still in view hierarchy)
// This handles the case where content changes and the old scroll view is unmounted
if (_pinnedScrollView && ![_pinnedScrollView isDescendantOfView:self]) {
  [self clearScrollable];
}

// Already set up with same inset and valid scroll view
if (_pinnedScrollView && _bottomInset == bottomInset) {
  return;
}
```

## Environment

- react-native-true-sheet: 3.8.1
- React Native: 0.81.5
- iOS: 18+
- New Architecture: Enabled

## Workaround

Using `patch-package` with the fix described above resolves the issue.

sbs44:
## Summary

Content gets clipped when dynamically switching sheet content (e.g., SearchView → SettingsView). Dragging the sheet and releasing fixes it, suggesting a stale reference.

`setupScrollable:bottomInset:` has an early return that doesn't verify `_pinnedScrollView` is still in the view hierarchy. When content changes, the old ScrollView is unmounted but we keep the reference.

This adds an `isDescendantOfView:` check to clear invalid references.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Tested with an app switching between two content views with ScrollViews. Content now renders correctly on first open.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

Fixes #481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed content clipping when sheet content changes dynamically by clearing stale ScrollView reference. ([#482](https://github.com/lodev09/react-native-true-sheet/pull/482) by [@sbs44](https://github.com/sbs44))
+
 ## 3.8.1
 
 ### 📖 Documentation
--- a/ios/TrueSheetContentView.mm
+++ b/ios/TrueSheetContentView.mm
@@ -93,7 +93,13 @@ using namespace facebook::react;
     return;
   }
 
-  // Already set up with same inset
+  // Check if pinned scroll view is still valid (still in view hierarchy)
+  // This handles the case where content changes and the old scroll view is unmounted
+  if (_pinnedScrollView && ![_pinnedScrollView isDescendantOfView:self]) {
+    [self clearScrollable];
+  }
+
+  // Already set up with same inset and valid scroll view
   if (_pinnedScrollView && _bottomInset == bottomInset) {
     return;
   }


lodev09:
:rocket: _This pull request is included in v3.8.2. See [Release 3.8.2](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.8.2) for release notes._

PR #499, opened 2026-02-06
refactor: use codegen enums for backgroundBlur, anchor, insetAdjustment

lodev09:
## Summary

Use codegen-generated enums instead of raw strings for `backgroundBlur`, `anchor`, and `insetAdjustment` props on both iOS and Android.

- **iOS**: Use C++ codegen enums directly in `.mm` files, `NSInteger` in headers
- **Android**: Kotlin enum classes with `fromString()` converters
- **TS spec**: Props now use proper string union types instead of plain strings

## Type of Change

- [x] New feature

## Test Plan

Built and tested on iOS and Android.

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web


lodev09:

PR #501, opened 2026-02-06
fix(android): use M3 baseline colors for default sheet background

lodev09:
## Summary

Use hardcoded M3 baseline `surfaceContainerLow` colors for the default sheet background instead of resolving from the app theme. The previous approach used `Theme.resolveAttribute(colorSurfaceContainerLow)` which fails silently for AppCompat-based apps (most RN apps), always falling back to white regardless of dark mode.

Now uses the correct M3 baseline values:
- Light: `#F7F2FA` (neutral96)
- Dark: `#1D1B20` (neutral10)

## Type of Change

- [x] Bug fix

## Test Plan

Tested on Android in dark mode — sheet now shows the correct dark surface color instead of white.

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web


lodev09:

PR #505, opened 2026-02-07
feat(ios): enable synchronous state update for RN 0.82+

lodev09:
## Summary

Enable synchronous state updates on iOS for React Native 0.82+ using `unstable_Immediate` update mode. This processes state updates in the same layout pass instead of asynchronously.

Note: This doesn't have a visible effect yet since we still rely on native layout constraints for container sizing. In the future, once the API becomes stable, we can drop native layout constraints entirely in favor of synchronous Yoga layout.

## Type of Change

- [x] New feature

## Test Plan

Tested on iOS with RN 0.82+.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### 🎉 New features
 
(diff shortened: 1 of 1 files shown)


lodev09:

PR #506, opened 2026-02-07
fix(ios): avoid deadlock when registering screens event observer

### LINKED ISSUE #504: iOS: Main thread deadlock (watchdog kill) when keyboard events fire while sheet is mounted

## Bug

The app gets killed by the iOS watchdog (`0x8BADF00D` — "exhausted real time allowance of 10.00 seconds") due to a main thread deadlock. Happens when a text input triggers keyboard events while a TrueSheet is mounted (Fabric/new arch).

## Root cause

Re-entrant lock upgrade deadlock in `RNScreensEventObserver.mm`:

1. `KeyboardControllerView` dispatches `onFocusedInputSelectionChanged`
2. React's `EventListenerContainer::willDispatchEvent()` acquires a **shared lock** on its internal `shared_mutex`
3. Reanimated's event listener processes the event, commits shadow tree updates
4. During mount instructions, `TrueSheetView updateState:oldState:` calls `[_screensEventObserver startObservingWithState:]`
5. `startObservingWithState:` calls `dispatcherPtr->addListener(_eventListener)` which needs an **exclusive lock** on the same `shared_mutex`
6. **Deadlock** — can't upgrade shared to exclusive on the same thread. The JS thread also gets blocked as collateral.

## Stack trace (Thread 0, main thread)

```
EventListenerContainer::willDispatchEvent()          <- shared lock acquired
  -> Reanimated event listener callback
    -> ReanimatedModuleProxy::handleRawEvent
      -> performOperations -> commitUpdates
        -> ShadowTree::commit -> tryCommit -> mount
          -> TrueSheetView updateState:oldState:
            -> RNScreensEventObserver startObservingWithState:
              -> EventListenerContainer::addListener()  <- needs exclusive lock = DEADLOCK
```

## Reproduction

1. Use a TrueSheet with a TextInput (e.g. in the footer)
2. Use `react-native-keyboard-controller` as the keyboard provider
3. Use `react-native-reanimated` (which registers an event listener)
4. Type in the text input or move the cursor
5. App freezes and gets killed after 10 seconds

## Environment

- react-native-true-sheet: 3.8.1 (also confirmed unfixed on main/v3.9.0-beta.1)
- react-native-keyboard-controller: 1.20.4
- react-native-reanimated: 4.2.1
- react-native-screens: 4.20.0
- React Native: 0.79 (Fabric/new arch)
- iOS Simulator, Expo 54+

## Suggested fix

Defer the `addListener` call in `startObservingWithState:` to the next runloop iteration:

```objc
// Instead of:
dispatcherPtr->addListener(_eventListener);

// Do:
auto listener = _eventListener;
auto dispatcher = dispatcherPtr;
dispatch_async(dispatch_get_main_queue(), ^{
  dispatcher->addListener(listener);
});
```

This breaks the synchronous chain so `addListener` runs after `willDispatchEvent` releases the shared lock. The one-tick delay is fine since the observer is just registering for future `topWillDisappear`/`topWillAppear` events.

Happy to open a PR with this fix.

### Affected Platforms

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Other

### Library Version

3.8.1

### Environment Info

```shell
Environment Info:                                                                                                         
  System:                                                                                                                 
    OS: macOS 26.2                                                                                                          
    CPU: (8) arm64 Apple M1 Pro                                                                                             
  Binaries:                                                                                                                 
    Node: 25.2.1                                                                                                            
    npm: 11.6.2                                                                                                             
  SDKs:                                                                                                                     
    iOS SDK: 26.2                                                                                                           
  IDEs:                                                                                                                     
    Xcode: 26.2/17C52                                                                                                       
  npmPackages:                                                                                                              
    react: 19.2.0
    react-native: 0.83.1
  iOS:
    hermesEnabled: true
    newArchEnabled: true

  react-native-true-sheet: 3.8.1
  react-native-keyboard-controller: 1.20.4
  react-native-reanimated: 4.2.1
  react-native-screens: 4.20.0
  Expo SDK: 54+
```

### Steps to Reproduce

  1. Wrap the app with KeyboardProvider from react-native-keyboard-controller
  2. Present a TrueSheet that contains a TextInput (e.g. in the footer prop)
  3. Tap the TextInput to focus it
  4. Type or move the cursor
  5. App freezes and gets killed by the iOS watchdog after 10 seconds (0x8BADF00D)

### Repro

https://github.com/lodev09/react-native-true-sheet

### Additional Context

hth!

lodev09:
## Summary

Move `startObservingWithState:` from `updateState:oldState:` to `presentAtIndex:` to avoid a main thread deadlock (`0x8BADF00D`) when keyboard events fire while a TrueSheet is mounted.

The deadlock occurs because `updateState:oldState:` can be called during mount instructions inside an event dispatch (which holds a shared lock on `EventListenerContainer`'s mutex), and `addListener` needs an exclusive lock on the same mutex.

Moving to `presentAtIndex:` is correct since the observer only needs to listen for screen lifecycle events while the sheet is presented.

Fixes #504

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Verified the screens event observer still registers correctly when presenting a sheet within a navigation stack.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - **iOS**: Fixed crash and blank sheet when using Reanimated exiting animations. ([#493](https://github.com/lodev09/react-native-true-sheet/pull/493) by [@lodev09](https://github.com/lodev09))
 - **Android**: Fixed default background color not respecting dark mode on AppCompat-based apps. ([#501](https://github.com/lodev09/react-native-true-sheet/pull/501) by [@lodev09](https://github.com/lodev09))
 - **Web**: Fixed default background color not respecting dark mode. ([#502](https://github.com/lodev09/react-native-true-sheet/pull/502) by [@lodev09](https://github.com/lodev09))
+- **iOS**: Fixed main thread deadlock when keyboard events fire while sheet is mounted. ([#506](https://github.com/lodev09/react-native-true-sheet/pull/506) by [@lodev09](https://github.com/lodev09))
 
 ### 💡 Others
 
--- a/ios/TrueSheetView.mm
+++ b/ios/TrueSheetView.mm
@@ -272,8 +272,6 @@ using namespace facebook::react;
   if (_controller) {
     [self updateStateWithSize:_controller.view.frame.size];
   }
-
-  [_screensEventObserver startObservingWithState:_state.get()->getData()];
 }
 
 /**
@@ -467,6 +465,7 @@ using namespace facebook::react;
   [_controller setupActiveDetentWithIndex:index];
 
   [_screensEventObserver capturePresenterScreenFromView:self];
+  [_screensEventObserver startObservingWithState:_state.get()->getData()];
 
   [presentingViewController presentViewController:_controller
                                          animated:animated


lodev09:

efstathiosntonas:
Hi @lodev09 , we recently hit deadlocks on ios and Opus 5 lead me here based on evidence from Sentry. 

It created this patch, not sure if it fixes the deadlocks, will release it in production and update here. 

using rn@0.83.10 Fabric. 

`@lodev09__react-native-true-sheet@3.11.9.patch`

```patch
diff --git a/ios/core/RNScreensEventObserver.mm b/ios/core/RNScreensEventObserver.mm
index dd3c374..153d1e3 100644
--- a/ios/core/RNScreensEventObserver.mm
+++ b/ios/core/RNScreensEventObserver.mm
@@ -65,6 +65,13 @@ using namespace facebook::react;
         return false;
       }
 
+      // Listener removal is deferred (see -stopObserving), so an event can still be
+      // delivered for one runloop tick after observing stopped. -stopObserving resets
+      // _eventDispatcher synchronously, so it doubles as the "still observing" flag.
+      if (strongSelf->_eventDispatcher.expired()) {
+        return false;
+      }
+
       if (auto family = event.shadowNodeFamily.lock()) {
         Tag screenTag = family->getTag();
         if (![strongSelf->_screenTags containsObject:@(screenTag)]) {
@@ -123,7 +130,22 @@ using namespace facebook::react;
       return false;
     });
 
-    dispatcherPtr->addListener(_eventListener);
+    // `startObservingWithState:` runs from -presentAtIndex:, which is reachable from
+    // -didMoveToWindow, i.e. from inside a mounting transaction. That transaction can be
+    // running inside `EventListenerContainer::willDispatchEvent`, which holds a *shared*
+    // lock on the dispatcher's shared_mutex. `addListener` takes the *exclusive* lock on
+    // the same mutex from the same thread, which deadlocks the main thread until the iOS
+    // watchdog kills the app (0x8BADF00D). Registering on the next runloop tick runs after
+    // the shared lock is released; the observer only needs future screen lifecycle events.
+    // Upstream: lodev09/react-native-true-sheet#504 (PR #506 moved the call site but kept
+    // the synchronous exclusive lock, so the deadlock survives via -didMoveToWindow).
+    auto weakDispatcher = _eventDispatcher;
+    auto listener = _eventListener;
+    dispatch_async(dispatch_get_main_queue(), ^{
+      if (auto dispatcher = weakDispatcher.lock()) {
+        dispatcher->addListener(listener);
+      }
+    });
   }
 }
 
@@ -131,9 +153,15 @@ using namespace facebook::react;
   [self stopInteractiveTracking];
 
   if (_eventListener) {
-    if (auto dispatcher = _eventDispatcher.lock()) {
-      dispatcher->removeListener(_eventListener);
-    }
+    // Same deadlock hazard as -startObservingWithState:: removeListener takes the exclusive
+    // lock, and unmount also runs inside a mounting transaction. Defer it the same way.
+    auto weakDispatcher = _eventDispatcher;
+    auto listener = _eventListener;
+    dispatch_async(dispatch_get_main_queue(), ^{
+      if (auto dispatcher = weakDispatcher.lock()) {
+        dispatcher->removeListener(listener);
+      }
+    });
     _eventListener = nullptr;
   }
   _eventDispatcher.reset();
```

PR #538, opened 2026-02-15
fix(android): apply scroll drag workaround for non-scrollable sheets with ScrollView

lodev09:
## Summary

Remove the `isScrollable` gate from `onInterceptTouchEvent` so the scroll drag workaround applies to all sheets containing a ScrollView whose content is shorter than the viewport — not just sheets with `scrollable` prop enabled.

## Type of Change

- [x] Bug fix

## Test Plan

- Open a non-scrollable sheet that contains a ScrollView with content shorter than the viewport
- Verify the sheet can be dragged up/down

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web


lodev09:

PR #545, opened 2026-02-16
fix(ios): custom grabber not centering on rotation

lodev09:
## Summary

Custom grabber view did not realign to center when the device rotated. The grabber frame was set manually in `applyConfiguration` without autoresizing masks, so UIKit had no way to adjust its position when the parent view resized.

Added `UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleRightMargin` so the grabber stays centered on rotation.

## Type of Change

- [x] Bug fix

## Test Plan

- Presented a sheet with custom grabber options
- Rotated the device — grabber stays centered

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -1,4 +1,4 @@
-name: Changelog
+name: Check Changelog
 on:
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


lodev09:
:rocket: _This pull request is included in v3.9.1. See [Release 3.9.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.9.1) for release notes._

PR #548, opened 2026-02-18
fix: use delegate method instead of modalInPresentation for non-dismissible sheets

### LINKED ISSUE #547: isModalInPresentation blocks touch passthrough for undimmed sheets
## Bug Description

When using `dismissible={false}` with `dimmed={false}`, touches in the area above the sheet (outside the sheet's bounds) are completely blocked on real iOS devices. This prevents interaction with the presenting view controller's content (map, buttons, search bars, etc.).

The issue occurs because `dismissible={false}` sets `isModalInPresentation = YES` on the view controller. Per [Apple's documentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/ismodalinpresentation):

> When you set this property to YES, UIKit **ignores events outside the view controller's bounds** and prevents interactive dismissal of the view controller while it is onscreen.

This conflicts with the `largestUndimmedDetentIdentifier` configuration, which is designed to allow touch passthrough for undimmed detents.

## Steps to Reproduce

1. Create a TrueSheet with `dismissible={false}`, `dimmed={false}`, and small detents (e.g., `[0.1, 0.29, 1]`)
2. Present the sheet
3. Try to tap buttons or interact with content behind the sheet on a **real iOS device**

## Expected Behavior

Touches above the sheet should pass through to the presenting view controller (since the sheet is undimmed via `largestUndimmedDetentIdentifier`).

## Actual Behavior

All touches above the sheet are dropped/ignored by UIKit. Only touches directly on the sheet content work.

**Note:** This works correctly in the iOS Simulator because the simulator's mouse-to-touch conversion bypasses some native event processing. It only fails on real devices.

## Proposed Fix

Replace `isModalInPresentation = YES` with delegate-based dismissal prevention using `presentationControllerShouldDismiss:`. This achieves the same non-dismissible behavior without blocking touch passthrough:

**TrueSheetViewController.h** — Add property:
```objc
@property (nonatomic, assign) BOOL dismissible;
```

**TrueSheetViewController.mm** — Add delegate method:
```objc
- (BOOL)presentationControllerShouldDismiss:(UIPresentationController *)presentationController {
  return self.dismissible;
}
```

**TrueSheetView.mm** — Replace:
```objc
// Before:
_controller.modalInPresentation = !newProps.dismissible;

// After:
_controller.dismissible = newProps.dismissible;
```

This way:
- `dismissible={false}` → delegate returns NO → interactive dismissal is prevented
- `isModalInPresentation` stays at its default (NO) → events outside the sheet are NOT ignored
- Touch passthrough works normally via `largestUndimmedDetentIdentifier`

## Environment

- TrueSheet version: 3.9.0
- React Native: 0.79 (New Architecture / Fabric)
- iOS: 18.x (real device)
- Xcode: 16.x

isaacrowntree:
## Problem

When `dismissible={false}`, TrueSheet sets `isModalInPresentation = YES` on the view controller. Per Apple's documentation:

> When you set this to YES, UIKit ignores events outside the view controller's bounds and prevents the interactive dismissal of the view controller while it is onscreen.

This causes **all touch events to be dropped** for content behind undimmed sheets on physical iOS devices — taps, pans, pinch-to-zoom, and all gestures are swallowed. The issue does not reproduce in the iOS Simulator because it uses a different touch delivery path.

### Reproduction

1. Present a TrueSheet with `dismissible={false}` and `dimmed={false}`
2. Set a small detent (e.g., `[0.1, 0.3, 1.0]`)
3. With the sheet at the smallest detent, try to interact with content above the sheet (tap buttons, pan a map, pinch-to-zoom)
4. **Expected:** Touches pass through to content behind the sheet
5. **Actual:** All touches above the sheet are blocked on physical devices

### Root Cause

`isModalInPresentation` is a blunt instrument — it prevents **both** interactive dismissal **and** touch passthrough. For undimmed sheets where you want to prevent swipe-to-dismiss but still allow interaction with background content, this property breaks the expected UX.

## Fix

Replace `isModalInPresentation` with the `presentationControllerShouldDismiss:` delegate method, which prevents interactive dismissal **without** blocking touch events outside the sheet's bounds.

### Changes

| File | Change |
|------|--------|
| `TrueSheetViewController.h` | Add `dismissible` property |
| `TrueSheetViewController.mm` | Implement `presentationControllerShouldDismiss:` delegate method |
| `TrueSheetView.mm` | Replace `modalInPresentation` assignment with `dismissible` property |

### Before (blocks all touches):
```objc
_controller.modalInPresentation = !newProps.dismissible;
```

### After (prevents dismiss without blocking touches):
```objc
// TrueSheetViewController.mm
- (BOOL)presentationControllerShouldDismiss:(UIPresentationController *)presentationController {
  return self.dismissible;
}

// TrueSheetView.mm
_controller.dismissible = newProps.dismissible;
```

## Testing

- Verified on physical iPhone 15 Pro (iOS 18) — touches pass through to content behind undimmed sheet at all detents
- No regression: `dismissible={false}` still prevents swipe-to-dismiss
- `dismissible={true}` still allows interactive dismissal

Fixes #547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed touch passthrough on undimmed sheets when `dismissible` is `false`. ([#548](https://github.com/lodev09/react-native-true-sheet/pull/548) by [@isaacrowntree](https://github.com/isaacrowntree))
+
 ### 💡 Others
 
 - Documented `auto` detent limitation when `scrollable` is enabled. ([#554](https://github.com/lodev09/react-native-true-sheet/pull/554) by [@lodev09](https://github.com/lodev09))


isaacrowntree:
Minimal reproduction repository: https://github.com/isaacrowntree/truesheet-touch-repro

Run on a physical iOS device to reproduce the touch-blocking behavior.

lodev09:
@isaacrowntree thanks. Looks good! Could you add entry to changelog please?

lodev09 (approved):

lodev09:
:rocket: _This pull request is included in v3.9.5. See [Release 3.9.5](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.9.5) for release notes._

PR #550, opened 2026-02-18
fix(ios): use window height instead of screen height for detent calculations

lodev09:
## Summary

Use `window.bounds.size.height` instead of `UIScreen.mainScreen.bounds.size.height` for detent calculations. On iPad with Stage Manager or Split View, the window height can differ from the full screen height, causing incorrect detent positioning.

## Type of Change

- [x] Bug fix

## Test Plan

- Tested on iPad with Stage Manager / Split View scenes
- Verified detent calculations use correct window height

## Checklist

- [x] I tested on iOS
- [ ] ~~I tested on Android~~ (iOS-only change)
- [ ] ~~I tested on Web~~ (iOS-only change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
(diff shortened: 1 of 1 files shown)


lodev09:
:rocket: _This pull request is included in v3.9.3. See [Release 3.9.3](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.9.3) for release notes._

PR #551, opened 2026-02-19
fix(ios): queue content size changes during presentation animation

lodev09:
## Summary

Content size changes arriving while the sheet is animating in (between `presentViewController:` and `viewDidAppear:`) were silently dropped. `setupSheetDetentsForSizeChange` guards on `isPresented`, which is only `YES` after `viewDidAppear:`. This caused sheets with `initialDetentIndex` to show stale detent heights when content changed during mount.

Now queues pending size changes via `_pendingSizeChange` flag and flushes them in `viewControllerDidPresentAtIndex:` after presentation completes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Use `initialDetentIndex={0}` with content that changes size shortly after mount
- Verify the sheet adjusts to the correct height after presentation completes
- Verify normal size changes while presented still work (no regression)

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed content size changes being dropped during presentation animation when using `initialDetentIndex`. ([#551](https://github.com/lodev09/react-native-true-sheet/pull/551) by [@lodev09](https://github.com/lodev09))
+
 ## 3.9.3
 
 ### 🐛 Bug fixes


lodev09:
:rocket: _This pull request is included in v3.9.4. See [Release 3.9.4](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.9.4) for release notes._

PR #557, opened 2026-02-21
fix(ios): queue detent updates during presentation animation

### LINKED ISSUE #555: Missing detent update when happening during presentation

### Bug Summary

We use onLayout to calculate the height of the first detent depending on which type of item we display in our sheet. Sometimes the onLayout reports the updated height right after the present-call has started the presentation which means that it presents for example 0.4 while we update the detent to 0.5. 

That it initially renders 0.4 is totally fine, but the bug is that we get stuck in a scenario where the sheet is opened at detent 0.4 even though it's update to 0.5 positionChange etc reports 0.5 as well even though we are at 0.4.

I was able to solve it by introducing another boolean called _didChangeDetentsWhilePresenting that when true in viewControllerDidPresentAtIndex applies the detent changes and updates the sheet but I am not sure if that it the solution you would go for.

### Affected Platforms

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Other

### Library Version

3.9.4

### Environment Info

```shell
System:
  OS: macOS 26.1
  CPU: (8) arm64 Apple M1
  Memory: 184.72 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.13.0
    path: /Users/viktorlindberg/.nvm/versions/node/v24.13.0/bin/node
  Yarn:
    version: 1.22.22
    path: /Users/viktorlindberg/.nvm/versions/node/v24.13.0/bin/yarn
  npm:
    version: 11.6.2
    path: /Users/viktorlindberg/.nvm/versions/node/v24.13.0/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.16.2
    path: /Users/viktorlindberg/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2025.2 AI-252.27397.103.2522.14617522
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.17
    path: /usr/bin/javac
  Ruby:
    version: 3.3.0
    path: /Users/viktorlindberg/.rbenv/shims/ruby
npmPackages:
    installed: 20.1.1
    wanted: ^20.1.1
  react:
    installed: 19.1.0
    wanted: 19.1.0
  react-native:
    installed: 0.81.5
    wanted: 0.81.5
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

### Steps to Reproduce

1. Update detent in a setTimeout with timeout 0 while and then directly call present on the sheet

In the videos below I update the detent to 0.5 while the presentation is ongoing. 
First video shows that the sheet gets stuck at 0.4 (even though positionChange etc reports 0.5).
Second video shows the sheet presenting at 0.4 but then instantly fixing it by going to 0.5 which is the new detent. 

### Repro

https://github.com/vilindberg/react-native-true-sheet-invalid-detent-update

### Additional Context

| Before fix | After fix |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/e63d2ddc-b337-4abc-8932-441e386f42f3" /> | <video src="https://github.com/user-attachments/assets/60cef038-4d54-4b2e-84d1-7eedf36ffa6b" /> |

lodev09:
## Summary

Fixes #555. When detents or content size changed during the sheet presentation animation, `UISheetPresentationController` would silently ignore the update, leaving the sheet stuck at the old detent while events reported the new value.

This PR defers prop/detent updates and content size changes that arrive while `isBeingPresented` is `YES`, then flushes them in `viewControllerDidPresentAtIndex:` once the animation completes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Present sheet with detent update triggered via `setTimeout` during animation — sheet now correctly animates to the updated detent after presentation completes
- Content size changes during presentation (`auto` detent) are queued and applied after present
- Detent/prop changes while already presented still apply immediately (no regression)

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 🐛 Bug fixes
 
 - **iOS**: Fixed touch passthrough on undimmed sheets when `dismissible` is `false`. ([#548](https://github.com/lodev09/react-native-true-sheet/pull/548) by [@isaacrowntree](https://github.com/isaacrowntree))
+- **iOS**: Fixed detent updates being ignored when changed during presentation animation. ([#557](https://github.com/lodev09/react-native-true-sheet/pull/557) by [@lodev09](https://github.com/lodev09))
 
 ### 💡 Others
 


lodev09:
:rocket: _This pull request is included in v3.9.5. See [Release 3.9.5](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.9.5) for release notes._

PR #589, opened 2026-03-04
fix: footer swallowing touch events

isaacrowntree:
## Summary

- **iOS**: Footer buttons/touchables don't respond to taps because Fabric's touch system hit-tests using stale Yoga frames while AutoLayout has moved the footer to a different position. Fix: give the footer its own `RCTSurfaceTouchHandler` so it has an independent coordinate space rooted at its actual frame.
- **Android**: Footer touches dispatched to JS twice — once by `dispatchTouchEvent` routing to footer's `RootView`, and again by `onInterceptTouchEvent`/`onTouchEvent` on the ViewController. Fix: skip JS dispatch in the ViewController when the touch is in footer bounds.
- **iOS**: Footer ignores safe area insets — content renders behind the home indicator on notched iPhones. Fix: pin to `safeAreaLayoutGuide.bottomAnchor` instead of `bottomAnchor`.
- **Android**: Footer renders behind the navigation/gesture bar. Fix: subtract `bottomInset` in `positionFooter` (skipped when keyboard is visible since `currentKeyboardInset` already accounts for it).

## Root cause (iOS touch bug)

`TrueSheetFooterView.updateLayoutMetrics` skips `[super updateLayoutMetrics:...]` after initial layout (line 93-96), so the Yoga frame goes stale while AutoLayout moves the view to the container bottom. `RCTSurfaceTouchHandler` on the container hit-tests using Yoga frames → can't find the footer at its visual position → touches dropped or delivered to the wrong component.

The fix mirrors the existing Android pattern where `TrueSheetFooterView` implements `RootView` with its own `JSTouchDispatcher` — on iOS we achieve the same by giving the footer its own `RCTSurfaceTouchHandler`.

## Test plan

- [ ] Tap `Button` and `TouchableOpacity` in footer on **physical iOS device** — taps register
- [ ] Tap footer buttons on **physical Android device** — taps register, no double-press
- [ ] Footer content doesn't render behind home indicator (notched iPhone)
- [ ] Footer content doesn't render behind gesture bar (Android)
- [ ] Keyboard avoidance still works — footer moves up with keyboard
- [ ] Test with `dismissible={false}` 
- [ ] Test at different detent sizes (small, medium, full)
- [ ] Test with and without `insetAdjustment="automatic"`

Reproduction app: https://github.com/isaacrowntree/truesheet-touch-repro

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed footer swallowing touch events by giving it a dedicated `RCTSurfaceTouchHandler`, bypassing stale Yoga frame hit-testing. ([#589](https://github.com/lodev09/react-native-true-sheet/pull/589) by [@isaacrowntree](https://github.com/isaacrowntree))
+- **Android**: Fixed double JS touch dispatch for footer touches — footer's own `RootView` now exclusively handles events in its bounds. ([#589](https://github.com/lodev09/react-native-true-sheet/pull/589) by [@isaacrowntree](https://github.com/isaacrowntree))
+
 ## 3.10.0
 
 ### 🎉 New features
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
@@ -1267,6 +1267,19 @@ class TrueSheetViewController(private val reactContext: ThemedReactContext) :
   // MARK: - RootView Touch Handling
   // =============================================================================
 
+  /**
+   * Check if a touch event is within the footer's screen bounds.
+   */
+  private fun isTouchInFooter(event: MotionEvent): Boolean {
+    val footer = containerView?.footerView ?: return false
+    if (!footer.isShown) return false
+    val loc = ScreenUtils.getScreenLocation(footer)
+    val x = event.rawX.toInt()
+    val y = event.rawY.toInt()
+    return x >= loc[0] && x <= loc[0] + footer.width &&
+      y >= loc[1] && y <= loc[1] + footer.height
+  }
+
   override fun dispatchTouchEvent(event: MotionEvent): Boolean {
     val footer = containerView?.footerView
     if (footer != null && footer.isShown) {
@@ -1293,17 +1306,24 @@ class TrueSheetViewController(private val reactContext: ThemedReactContext) :
   }
 
   override fun onInterceptTouchEvent(event: MotionEvent): Boolean {
-    eventDispatcher?.let {
-      jsTouchDispatcher.handleTouchEvent(event, it, reactContext)
-      jsPointerDispatcher.handleMotionEvent(event, it, true)
+    // Skip JS dispatch for footer touches — the footer's own RootView handles them.
+    // This prevents the same touch event being dispatched to JS twice.
+    if (!isTouchInFooter(event)) {
+      eventDispatcher?.let {
+        jsTouchDispatcher.handleTouchEvent(event, it, reactContext)
+        jsPointerDispatcher.handleMotionEvent(event, it, true)
+      }
     }
     return super.onInterceptTouchEvent(event)
   }
 
   override fun onTouchEvent(event: MotionEvent): Boolean {
-    eventDispatcher?.let {
-      jsTouchDispatcher.handleTouchEvent(event, it, reactContext)
-      jsPointerDispatcher.handleMotionEvent(event, it, false)
+    // Skip JS dispatch for footer touches — handled by footer's own RootView.
+    if (!isTouchInFooter(event)) {
+      eventDispatcher?.let {
+        jsTouchDispatcher.handleTouchEvent(event, it, reactContext)
+        jsPointerDispatcher.handleMotionEvent(event, it, false)
+      }
     }
     super.onTouchEvent(event)
     return true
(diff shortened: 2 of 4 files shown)


lodev09:
Thanks @isaacrowntree!

Do you have a repro from a fork of this repo that I can replicate on my end prior to your changes? I want to test this properly.

isaacrowntree:
Hey @lodev09! Here's a repro branch on my fork of this repo:

**[`isaacrowntree/react-native-true-sheet@repro/footer-touch`](https://github.com/isaacrowntree/react-native-true-sheet/tree/repro/footer-touch)**

It adds a `FooterReproSheet` to the **Test Screen** in the example app with three variants covering all the test cases:

| Button | What it tests |
|--------|--------------|
| **Footer Repro** | Tappable button + TextInput in footer (touch bug + keyboard avoidance) |
| **Footer (dismissible=false)** | Same footer with `dismissible={false}` |
| **Footer (insetAdjustment=auto)** | Same footer with `insetAdjustment="automatic"` |

All three use `detents={['auto', 'medium', 'large']}` so you can test at different sizes.

On a physical device before the fix:
- **iOS**: footer button taps are silently dropped, TextInput can't be focused
- **Android**: footer button taps fire twice (double alert)
- **Both**: footer renders behind home indicator / gesture bar on notched devices

The existing `BasicSheet` footer is also made tappable for a quick sanity check.

I also rebased this PR on latest `main` so the conflicts are resolved.

lowbits:
facing the same issue on iPhone Xs

lodev09:
@isaacrowntree I tested this with your provided repro and I couldn't replicate. what device are you testing this with? Also the footer is crashing

isaacrowntree:
Hey @lodev09! I've updated the repro branch and also simplified it to make reproduction much easier.

### Updated repro branch

**[`isaacrowntree/react-native-true-sheet@repro/footer-touch`](https://github.com/isaacrowntree/react-native-true-sheet/tree/repro/footer-touch)**

Changes:
- **Rebased onto latest `main`** (`3.10.0-beta.3`) — resolves any stale native code issues
- **Bumped RN 0.82 → 0.84.1** + reanimated 4.3, worklets 0.8.1, screens 4.24 — builds cleanly on Xcode 26
- **Fixed TS errors** from beta.3 type changes (`SheetDetent`, `InsetAdjustment`, reanimated return types)
- **Replaced the example app** with a single-screen repro that auto-presents the sheet — no navigation needed

### How to test

```bash
git clone https://github.com/isaacrowntree/react-native-true-sheet.git --branch repro/footer-touch
cd react-native-true-sheet
yarn install
cd example/bare/ios && pod install && cd ../../..
# Update DEVELOPMENT_TEAM in Xcode to your signing team
# Build and run on a physical iOS device
```

The app auto-presents a sheet with two tappable buttons:

| Button | Location | Expected |
|--------|----------|----------|
| **CONTENT: 0/10** | Sheet body (control) | Always registers — should reach 10/10 |
| **FOOTER: 0/10** | Footer | **Bug: every 2nd tap is silently swallowed** |

Both buttons show a counter out of 10, turn **red** while counting and **green** at 10/10. The footer button also fires an `Alert` on each tap so you can clearly see which taps register and which are dropped.

### What I observed

- **Device:** iPhone 15 Pro, iOS 18
- **Content button:** 10 taps → 10/10 ✅ every time
- **Footer button:** 10 taps → ~5/10 ❌ — every 2nd tap is swallowed
- The pattern is very consistent: tap-register-tap-drop-tap-register-tap-drop

### Re: the crash you mentioned

Could you share the crash log / stack trace? With the updated branch (rebased on beta.3, RN 0.84) the crash may be resolved. The previous branch was pinned to beta.2 which could have had stale native state if your local pods came from a newer version.

lodev09:
Thanks @isaacrowntree.. I was able to reproduce. Please rebase and then remove the inset logic for now -- I only want to fix the touch issue, the inset issue is different and needs more thought and should be in a separate PR.

lodev09:
Also, the crash is still happening on this branch -- happens when I press the footer (in PromptSheet) and then dismissed.

> #12	0x0000000108d14668 in -[TrueSheetFooterView prepareForRecycle] at /Users/lodev09/Projects/labs/react-native-true-sheet/ios/TrueSheetFooterView.mm:111

lodev09 (changes requested):
See comments above

lodev09:
@isaacrowntree it's odd I can no longer replicate this when trying in main. Could you cleanup your repro and rebase in main? There's a lot of changes there that's adding noice and inconsistency with testing.

isaacrowntree:
Rebased on latest `main` (3.10.0) and scoped this PR down to just the touch fix per your request.

**Changes in this revision:**
- Dropped all safe-area / inset changes — split out to #644.
- Fixed the `prepareForRecycle` crash at `TrueSheetFooterView.mm:111`. Root cause was a double-detach: `didMoveToSuperview:nil` was already calling `detachFromView:` before `prepareForRecycle` ran, and `RCTSurfaceTouchHandler` asserts on detach-when-not-attached. Fixed by guarding both attach and detach with a `BOOL _footerTouchHandlerAttached` flag so either call path is idempotent.
- Preserved the `TrueSheetFooterViewDelegate` / `footerViewDidChangeSize` code that the old rebase had accidentally dropped.

**Repro branch cleaned up:** [`repro/footer-touch`](https://github.com/isaacrowntree/react-native-true-sheet/tree/repro/footer-touch) is now a single commit on top of `main` that only touches `example/bare/src/App.tsx` — no RN / reanimated / screens bumps, no other noise. Auto-presents the sheet with a CONTENT vs FOOTER tap counter (tap each to 10/10 — on affected builds the footer counter drops every second tap).

extradosages:
Tragic that this was closed! I think my team has encountered this very same bug!

lowbits:
i also rewrote my code, so its not using footer anymore...

PR #593, opened 2026-03-06
fix(android,ios): dead state after rapid present/dismiss

### LINKED ISSUE #590: Dead state after rapid present/dismiss — lazy loading race in onDidDismiss
## Description

After a few rapid open/close cycles, the sheet stops opening for the rest of the session. `present()` resolves without error but the sheet never appears. The only recovery is remounting the TrueSheet component (or restarting the app).

Affects **both Android and iOS** since the root cause is in the JS-side `TrueSheet.tsx` (shared layer).

We implemented app-side serialization as suggested in #573 (blocking present during closing, queueing, cooldowns). That fixed the rapid-tap race. But there's a separate issue in TrueSheet's lazy loading that app-side guards can't prevent.

## Impact

- `present()` resolves successfully — the app has no way to detect the failure
- The state is permanent until the component unmounts and remounts
- All subsequent `present()` calls silently no-op

## Root Cause — Lazy Loading Race

The race condition in `TrueSheet.tsx`:

1. Sheet is open, user swipe-dismisses
2. Native side runs cleanup → nulls internal sheet components (`coordinatorLayout`/`sheetView` on Android, equivalent on iOS)
3. Before JS `onDidDismiss` callback arrives, app calls `present()` → sets `this.isPresenting = true` (line 339)
4. `onDidDismiss` fires on JS side — sees `isPresenting === true` → **skips** `setState({ shouldRenderNativeView: false })` (line 274)
5. Next `present()` sees `shouldRenderNativeView` is still `true` → skips the lazy-load remount wait
6. Calls `presentByRef` → native finds null sheet components → "sheet is not ready" → resolves the promise silently
7. Native view is still in the React tree but the native sheet was destroyed by cleanup. All future `present()` calls hit "sheet is not ready" and silently resolve.

The core issue: `isPresenting` in `onDidDismiss` (line 274) prevents the `shouldRenderNativeView = false` reset needed to trigger a fresh mount cycle on the next `present()`.

## Why App-Side Guards Can't Prevent This

Even with lifecycle serialization, this race can still be hit because:
- The window between native `cleanupSheet()` and JS `onDidDismiss` is a bridge latency gap the app can't observe
- Once entered, `present()` resolves successfully — the app cannot detect the failure
- The "sheet is not ready" warning is logged natively but not surfaced to JS

## Suggested Fixes

**Option A:** In `present()`, check whether native sheet components exist before skipping the lazy-load mount wait. If `shouldRenderNativeView` is `true` but the native sheet was cleaned up, force a remount cycle.

**Option B:** In `onDidDismiss`, always set `shouldRenderNativeView = false` regardless of `isPresenting`. The in-flight present will trigger a remount via the existing lazy-load path.

**Option C:** Have native `presentByRef` reject the promise (instead of resolving) when sheet components are null, so the app can detect and recover.

## Environment

- react-native-true-sheet: 3.8.2
- React Native: 0.83
- Expo SDK: 55
- New Architecture / Fabric: enabled
- Platforms: Android and iOS

## Steps to Reproduce

1. Render a TrueSheet with lazy loading (default behavior)
2. Rapidly open and close the sheet (tap to open → swipe dismiss, repeat)
3. After a few cycles, the sheet stops opening
4. `present()` resolves without error but the sheet doesn't appear
5. All subsequent `present()` calls also silently fail until the component remounts

lodev09:
## Summary

Fixes dead state after rapid present/dismiss cycles where `present()` resolves without error but the sheet never appears.

- **Android**: Moved `createSheet()` and `setupScrollable()` from mount time (`addView`) to present time (`present()`), removing the early-return guard that checked for null sheet components
- **iOS**: Mirrored the Android pattern — moved scrollable setup from `mountChildComponentView` to `presentAtIndex:`, added `setupScrollable` helper for consistent usage across the view
- **Example**: Added stress test (long press "TrueSheet View" button) for rapid present/dismiss cycles

Closes #590

## Type of Change

- [x] Bug fix

## Test Plan

- Long press the "TrueSheet View" button in the example app to trigger 5 rapid present/dismiss cycles
- Verify sheet presents correctly after the stress test
- Verify normal present/dismiss still works
- Verify ScrollView sheets still scroll correctly

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### 🐛 Bug fixes
 
 - **iOS**: Fixed keyboard scroll positioning when sheet auto-expands from a smaller detent. ([#592](https://github.com/lodev09/react-native-true-sheet/pull/592) by [@lodev09](https://github.com/lodev09))
+- **Android**: Fixed dead state after rapid present/dismiss cycles. ([#593](https://github.com/lodev09/react-native-true-sheet/pull/593) by [@lodev09](https://github.com/lodev09))
 - **iOS**: Fixed position change not emitting when detent or index changed. ([#584](https://github.com/lodev09/react-native-true-sheet/pull/584) by [@lodev09](https://github.com/lodev09))
 - **Android**: Use RN `BackHandler` for back press detection for reliability across Android versions. ([#580](https://github.com/lodev09/react-native-true-sheet/pull/580) by [@lodev09](https://github.com/lodev09))
 
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetView.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetView.kt
@@ -119,9 +119,6 @@ class TrueSheetView(private val reactContext: ThemedReactContext) :
 
     if (child is TrueSheetContainerView) {
       child.delegate = this
-      viewController.createSheet()
-      setupScrollable()
-
       val surfaceId = UIManagerHelper.getSurfaceId(this)
       eventDispatcher?.dispatchEvent(MountEvent(surfaceId, id))
     }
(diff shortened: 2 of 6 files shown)


lodev09:
:rocket: _This pull request is included in v3.10.0. See [Release 3.10.0](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.10.0) for release notes._

PR #598, opened 2026-03-10
fix: macOS Designed for iPhone — liquid glass background + mouse drag support

isaacrowntree:
## Summary

Two fixes for macOS **"Designed for iPhone/iPad"** compatibility mode (Apple Silicon):

### 1. Liquid glass background

`sheet.backgroundEffect` with `UIColorEffect` is silently ignored by the macOS compatibility layer. This causes the default iOS 26 liquid glass material to show through instead of the configured `backgroundColor`. The same binary works correctly on real iOS devices (tested iPhone 17 Pro, iOS 26.2).

**Fix:** Remove the early `return` when `NSProcessInfo.processInfo.isiOSAppOnMac` is `YES`, so `view.backgroundColor` is also set as a fallback alongside `backgroundEffect`.

### 2. Mouse drag gestures

`UISheetPresentationController`'s built-in pan gesture doesn't respond to mouse/trackpad input on Mac. Users can only interact with the sheet by tapping the grabber.

**Fix:** Add a custom `UIPanGestureRecognizer` on the `presentedView` (guarded by `isiOSAppOnMac`) that computes a projected landing position on release using the drag's cumulative translation + velocity (decelerating over ~0.3s), then snaps to the nearest detent. This mimics native iOS release-to-snap behavior.

- Gesture covers the header/grabber area — scroll view content remains scrollable at any detent
- Uses `resolvedHeightForIndex:` (now exposed in `TrueSheetDetentCalculator.h`) for accurate detent position calculations
- No mid-drag snapping = no jitter
- Both fixes are gated behind `NSProcessInfo.processInfo.isiOSAppOnMac` (iOS 14+) — zero change on real iOS devices

## Test plan

- [ ] Run on **macOS "Designed for iPhone/iPad"** — sheet has solid `backgroundColor`, no liquid glass
- [ ] Run on **macOS** — drag header/grabber area up/down with mouse, release to snap to nearest detent
- [ ] Run on **macOS** — fast flick up/down snaps in the flick direction
- [ ] Run on **macOS** — scroll view content scrolls normally at all detents
- [ ] Run on **physical iOS device (iOS 26+)** — sheet background and drag unchanged
- [ ] Run on **iOS Simulator** — no regression
- [ ] Test with `backgroundColor` set
- [ ] Test with `backgroundBlur` set (no `backgroundColor`)
- [ ] Test with `dismissible={false}` and `dismissible={true}`

Reproduction app: https://github.com/isaacrowntree/truesheet-macos-liquid-glass-repro

### DIFF
--- a/ios/TrueSheetViewController.mm
+++ b/ios/TrueSheetViewController.mm
@@ -853,7 +853,7 @@ static BOOL TrueSheetPositionStateEquals(TrueSheetPositionState a, TrueSheetPosi
 
 #if RNTS_IPHONE_OS_VERSION_AVAILABLE(26_1)
   if (@available(iOS 26.1, *)) {
-    if (!self.isDesignCompatibilityMode) {
+    if (!self.isDesignCompatibilityMode && [self.sheet respondsToSelector:@selector(setBackgroundEffect:)]) {
       if (self.backgroundColor) {
         self.sheet.backgroundEffect = [UIColorEffect effectWithColor:self.backgroundColor];
       } else if (hasBlur) {


lodev09:
I'm not confident hacking our way to make this work on mac. I'd rather wait for apple to officially support this natively.
Closing this.

PR #604, opened 2026-03-15
fix(ios): animate sheet back to detent on keyboard dismiss

TheOmran:
## Bug

When using custom detents (especially `'auto'`), dismissing the keyboard causes the sheet to **snap back** to its detent position instead of smoothly animating the transition. This affects any sheet with text inputs that uses custom or auto-sized detents.

## Root Cause

`TrueSheetKeyboardObserver` checks `isFirstResponderWithinSheet` before forwarding `keyboardWillChangeFrame:` notifications to delegates. When the keyboard **hides**, the first responder has already resigned by the time the notification fires, so the check returns `NO` and the hide event is silently dropped.

Without the hide event being forwarded, `UISheetPresentationController` recalculates the custom detent position without being wrapped in an `animateChanges:` block, causing the snap instead of a smooth animation.

## Fix

1. **Track keyboard state per sheet** — Added `_wasShowingForSheet` flag to `TrueSheetKeyboardObserver`. When the keyboard shows for this sheet, the flag is set. On keyboard hide, the first-responder check is skipped if the flag was set (then reset), ensuring the hide event is always forwarded when the keyboard was previously shown for this sheet.

2. **Forward keyboard hide to the sheet** — Registered `TrueSheetContainerView` as a keyboard observer delegate. On `keyboardWillHide:`, it notifies the delegate via a new optional `containerViewKeyboardWillHide` method.

3. **Force animated detent reconfiguration** — `TrueSheetView` implements `containerViewKeyboardWillHide` by calling `setupSheetDetentsForSizeChange`, which wraps the detent reconfiguration in `[sheet animateChanges:]`, forcing UIKit to animate the transition back to the custom detent.

## Videos

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/62f0bea1-d216-4de3-aa2c-81e13f37349f" width="200" /> | <video src="https://github.com/user-attachments/assets/8b5811fa-3085-4fca-bb28-56f44d457f4f" width="200" /> |

## Files Changed

- `ios/core/TrueSheetKeyboardObserver.mm` — State tracking for keyboard show/hide per sheet
- `ios/TrueSheetContainerView.h` — New optional delegate method
- `ios/TrueSheetContainerView.mm` — Keyboard observer delegate implementation
- `ios/TrueSheetView.mm` — Animated detent reconfiguration on keyboard hide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - **Android**: Fixed dead state after rapid present/dismiss cycles. ([#593](https://github.com/lodev09/react-native-true-sheet/pull/593) by [@lodev09](https://github.com/lodev09))
 - **iOS**: Fixed position change not emitting when detent or index changed. ([#584](https://github.com/lodev09/react-native-true-sheet/pull/584) by [@lodev09](https://github.com/lodev09))
 - **Android**: Use RN `BackHandler` for back press detection for reliability across Android versions. ([#580](https://github.com/lodev09/react-native-true-sheet/pull/580) by [@lodev09](https://github.com/lodev09))
+- **iOS**: Fixed sheet snapping instead of animating back to detent on keyboard dismiss. ([#604](https://github.com/lodev09/react-native-true-sheet/pull/604) by [@TheOmran](https://github.com/TheOmran))
 
 ### 💡 Others
 


TheOmran:
@lodev09 🙏🏼

lodev09:
@TheOmran do you have a video showing before/after?

blixky:
Somebody pleaseeee merge this pr?

TheOmran:
> @TheOmran do you have a video showing before/after?

Updated PR description with videos! 🙂

lodev09:
@TheOmran I can't seem to replicate it. Are you able to in the repo example?

lodev09:
so I was able to replicate this after upgrading example to RN 0.83 (expo 55). This has something to do with RN internals and I'm investigating the cause. The animate dismiss behavior should be the default in UIKit, however RN is causing it to fail for some reason.

lodev09:
Okay I found it. Here's the related issue: https://github.com/facebook/react-native/issues/55116
PR or patch to fix: https://github.com/facebook/react-native/pull/55908

I'll just update troubleshooting in a separate PR for this since this is RN issue

lodev09:
https://github.com/lodev09/react-native-true-sheet/pull/632

PR #609, opened 2026-03-19
fix(ios): clean up gesture recognizer targets on sheet dismiss

davidecallegaro:
## Summary

- `setupGestureRecognizer` adds `handlePanGesture:` as a target to the presented view's and inner ScrollView's pan gesture recognizers via `addTarget:action:`, but there is no corresponding `removeTarget:action:` when the sheet disappears
- This leaves orphaned gesture handler targets that block scroll events on the presenting view controller when navigating away and returning (e.g. in a React Navigation stack)

## Changes

- **GestureUtil**: Added `detachPanGestureHandler:target:selector:` to mirror `attachPanGestureHandler`
- **TrueSheetViewController**: Added `cleanupGestureRecognizer` method
  - Called in `viewWillDisappear:` to remove targets before the view disappears
  - Re-attaches in `viewDidAppear:` when the sheet reappears (e.g. after a child sheet is dismissed)

## Reproduction

1. Present a TrueSheet with `scrollable={true}` on a screen
2. Navigate to another screen (push) while the sheet is presented
3. Navigate back
4. The presenting screen's ScrollView is frozen — cannot scroll, but buttons still work

## Test plan

- [x] Open sheet → navigate away → return: scroll works
- [x] Open sheet → present child sheet → dismiss child: parent sheet drag still works
- [x] Open sheet → dismiss normally: no regression


lodev09:
I tried looking into this more and I couldn't replicate the issue you are trying to fix here. Create an issue with a repro first then a PR that fixes, I'll confirm if this is actually an issue or something else is going on.

Closing

PR #620, opened 2026-03-30
Using dedent 'auto' does not take into account footer height

seanpcoyle:
## Summary

When using a dedent of 'auto', the header's height is taken into account but the footer's height is not. This pull request makes it such that the footer height is also accounted for.

The attached screenshots were captured using the basic sheet with an obnoxiously red footer. The most important part of this is to use a significantly tall footer to reproduce the issue.

The 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Tested by opening iOS, Android, and web basic sheets both before and after.

### Before
Simulator Screenshot - iPhone 17 Pro - 2026-03-29 at 21 02 32

### After (iOS)
Simulator Screenshot - iPhone 17 Pro - 2026-03-29 at 21 07 06

### After (Web)
image

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [x] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- Fixed dedent 'auto' not respecting footer height ([#620](https://github.com/lodev09/react-native-true-sheet/pull/620) by [@seanpcoyle](https://github.com/seanpcoyle))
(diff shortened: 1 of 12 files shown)


lodev09:
Thanks @seanpcoyle, see this related discussion for the footer https://github.com/lodev09/react-native-true-sheet/discussions/613

Footer is supposed to be floating so it has to be an option for users, not enforced.

seanpcoyle:
> Thanks @seanpcoyle, see this related discussion for the footer https://github.com/lodev09/react-native-true-sheet/discussions/613
> 
> 
> 
> Footer is supposed to be floating so it has to be an option for users, not enforced.

Ah, I had checked issues but not discussions. Thanks for pointing that out!

PR #644, opened 2026-04-10
fix: footer ignoring safe area bottom inset

isaacrowntree:
## Summary
Split out from #589 per maintainer request — #589 now contains only the touch-handling fix.

Two related fixes for footer content rendering under the system safe-area chrome on notched / gesture-bar devices:

- **iOS** (`TrueSheetFooterView.mm`): pin the footer's bottom constraint to `parentView.safeAreaLayoutGuide.bottomAnchor` instead of `parentView.bottomAnchor`, so footer content sits above the home indicator. When the keyboard is visible, its height already includes the safe-area inset, so the existing keyboard offset handling still works unchanged.
- **Android** (`TrueSheetViewController.positionFooter`): subtract `bottomInset` from `footerY` when the keyboard is not visible, so footer content sits above the navigation/gesture bar. When the keyboard is visible, `currentKeyboardInset` already accounts for the bottom inset, so we skip the extra offset to avoid double-compensation.

## Repro
Run `example/` on a notched device (iPhone 15 Pro or a Pixel with gesture nav), open `BasicSheet`, observe footer content clipped behind the home indicator / gesture bar. Screenshots to follow.

## Test plan
- [ ] iOS on notched device: footer content sits above the home indicator at all detents
- [ ] iOS with keyboard visible: footer rides the keyboard without a gap or overlap
- [ ] Android on gesture-nav device: footer content sits above the gesture bar
- [ ] Android with keyboard visible: footer rides the keyboard without a gap or overlap
- [ ] No regression in `BasicSheet` / `PromptSheet` example screens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed footer ignoring safe area insets by pinning to `safeAreaLayoutGuide.bottomAnchor` instead of `bottomAnchor`. ([#644](https://github.com/lodev09/react-native-true-sheet/pull/644) by [@isaacrowntree](https://github.com/isaacrowntree))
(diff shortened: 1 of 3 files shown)


lodev09:
@isaacrowntree you're making this as default behavior on footers.. how about those users who want to really render the footer behind the home indicator?

lodev09:
Closing this. Users can just adjust the footer position via insets in JS if needed.

PR #662, opened 2026-04-23
fix(ios): render dim in presenter view when presenter VC is detached

### LINKED ISSUE #661: iOS dim renders in wrong UIWindow when sheet is presented from a detached UIViewController

### Bug Summary

On iOS, when a `TrueSheet` is presented from a `UIViewController` that is not reachable from the key window's root VC via the `parent` / `presentingViewController` chain (i.e. "detached" in UIKit's terms), the sheet content itself renders correctly, but UIKit's `UISheetPresentationController` dim layer ends up in the wrong `UIWindow`. The net visual effect is that the sheet is visible but appears over undimmed content, while the dim lives behind whatever UIKit chose as the fallback container.

iOS itself surfaces this at runtime:

```
[UIKitCore] Presenting view controller <TrueSheetViewController: …> from detached view controller <RCTFabricModalHostViewController: …> is not supported, and may result in incorrect safe area insets and a corrupt root presentation. Make sure <RCTFabricModalHostViewController: …> is in the view controller hierarchy before presenting from it. Will become a hard exception in a future release.
```

### Affected Platforms

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Other

### Library Version

3.10.0

### Environment Info

```shell
System:
  OS: macOS 26.4.1
  CPU: (16) arm64 Apple M3 Max
  Memory: 2.03 GB / 64.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.12.0
    path: /Users/joel.kanzelmeyer/.nvm/versions/node/v24.12.0/bin/node
  Yarn:
    version: 4.12.0
    path: /Users/joel.kanzelmeyer/.nvm/versions/node/v24.12.0/bin/yarn
  npm:
    version: 11.6.2
    path: /Users/joel.kanzelmeyer/.nvm/versions/node/v24.12.0/bin/npm
  Watchman:
    version: 2024.10.14.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.15.2
    path: /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2025.3 AI-253.31033.145.2533.15113396
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.11
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
    installed: 20.1.3
    wanted: ^20.1.3
  react:
    installed: 19.1.0
    wanted: 19.1.0
  react-native:
    installed: 0.81.5
    wanted: 0.81.5
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

### Steps to Reproduce

The bug is triggered whenever TrueSheet's presenter is a `UIViewController` that cannot be walked back to the window's root VC via `parent` → `presenting`. Two common shapes that produce it:

1. **RN `<Modal>` inside a custom React Navigation navigator / pager:** the `<Modal>`'s presenter (`[self reactViewController]` in Fabric's `RCTModalHostViewComponentView`) is a VC attached to the view tree via `addSubview:` rather than `addChildViewController:` — detached from UIKit's containment. A TrueSheet rendered inside that Modal inherits the detachment.

2. **Any custom native/JS navigator** that parks its screens' VCs in the view hierarchy via `addSubview:` rather than proper VC containment.

Minimal JS repro shape (inside the non-trivial case):

```tsx
// inside a react-navigation screen (or similar), under New Arch
<Modal transparent visible>
  <View style={{ flex: 1 }}>
    <TrueSheet ref={ref} detents={[0.5]} dimmed>
      <Text>Hello</Text>
    </TrueSheet>
    <Button title="open" onPress={() => ref.current?.present()} />
  </View>
</Modal>
```

Observed vs expected

  | Observed | Expected
-- | -- | --
Sheet content | ✓ renders above everything | ✓
Dim layer | Appears in the key window behind the outer Modal's content | Covers the content directly below the sheet
Console | UIKit "detached view controller" warning | No warning

### Repro

https://github.com/kanzelm3/truesheet-detached-vc-repro

### Additional Context

_No response_

kanzelm3:
## Summary

Fixes #661.

On iOS, when a `TrueSheet` is presented from a `UIViewController` that is not reachable from the key window's root VC via the `parent` / `presentingViewController` chain (i.e. "detached" in UIKit's terms), the sheet content itself renders correctly, but UIKit's `UISheetPresentationController` dim layer ends up in the wrong `UIWindow`. iOS surfaces this at runtime with:

```
[UIKitCore] Presenting view controller <TrueSheetViewController: …> from detached
view controller <...> is not supported, and may result in incorrect safe area
insets and a corrupt root presentation ... Will become a hard exception in a future release.
```

Two changes, both in `ios/TrueSheetView.mm`:

1. **`findPresentingViewController` walks the responder chain** to locate the nearest ancestor VC. This correctly finds VCs parked in the view tree by custom navigators (which may use `addSubview:` rather than modal presentation), where the previous `window.rootViewController.presentedViewController` chain walk would miss them and fall back to the app root.

2. **New `isPresenterDetached:` helper** walks the `parent` → `presenting` chain and only returns NO if it reaches the window's root VC. Detachment is *inherited* — a VC can have a non-nil `presentingViewController` but still be detached because that presenter is itself detached, so a single-level check is not sufficient. When the presenter is detached AND `dimmed` is YES, a fallback dim `UIView` is added as a subview of the presenter's view so it lives in the correct window below the natively-presented sheet. The dim is removed (animated) in `dismissAnimated:`, `viewControllerWillDismiss`, `prepareForRecycle`, and `dealloc` so every dismissal path — including swipe-dismiss and Fabric view recycling — cleans up correctly.

No behavior change for the attached-presenter path.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Reproduced the bug against the issue's repro (RN `<Modal>` inside a `UIPageViewController`-backed tree, TrueSheet inside the Modal). Confirmed the dim now renders correctly covering the Modal content and the sheet animates in/out as expected.
- Verified swipe-dismiss and programmatic dismiss both clean up the fallback dim in lockstep with the sheet animation — no leaked views or frames.
- Verified that when the presenter is attached (normal case), `isPresenterDetached:` returns `NO` and no fallback dim is added; UIKit's native dim continues to work exactly as before.
- `yarn tidy` ✅
- `yarn test` ✅ (43 passed)
- Tested on iOS 26 simulator, RN 0.81, New Architecture enabled.

## Screenshots / Videos

Before (dim lands in wrong window, behind the `<Modal>` content):

_[happy to add a short screen recording from the repro app if helpful — let me know]_

After (dim correctly covers the presenter's view):

_[ditto]_

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android _(iOS-only change; Android unaffected)_
- [ ] I tested on Web _(iOS-only change)_
- [ ] I updated the documentation _(no API change)_
- [x] I added a changelog entry

## Notes for reviewer

- Happy to split this into two commits / PRs if you'd prefer to land the responder-walk separately from the fallback-dim — the former is the smaller, uncontroversial fix and the latter is the more opinionated one.
- The fallback dim currently always shows full opacity when the sheet is presented (matching `dimmedDetentIndex: 0` semantics, which is the common case). If you'd like it to respect `dimmedDetentIndex > 0` and fade with detent changes, happy to follow up — want to keep this PR focused for easier review.
- Dim color is `rgba(0, 0, 0, 0.4)` — close to UIKit's default. Open to changing if there's a preferred value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐛 Bug fixes
 
+- **iOS**: Fixed sheet dim rendering in the wrong `UIWindow` when the presenter view controller is detached from the window's VC hierarchy (e.g. inside a custom navigator or RN `<Modal>`). `findPresentingViewController` now walks the responder chain to find the real owning VC, and a fallback dim view is added to the presenter's view when the presenter is detached. ([#662](https://github.com/lodev09/react-native-true-sheet/pull/662) by [@kanzelm3](https://github.com/kanzelm3))
 - **Android**: Fixed focused input in sheet causing auto-focus on main screen input after dismiss. ([#649](https://github.com/lodev09/react-native-true-sheet/pull/649) by [@lodev09](https://github.com/lodev09))
 
 ## 3.10.0
--- a/ios/TrueSheetView.mm
+++ b/ios/TrueSheetView.mm
@@ -65,6 +65,10 @@ using namespace facebook::react;
   BOOL _pendingPropsUpdate;
   NSArray *_pendingDetents;
   RNScreensEventObserver *_screensEventObserver;
+  // Fallback dim view used when the presenter VC is detached from the window's
+  // VC hierarchy — UIKit's built-in sheet dim renders in the wrong UIWindow in
+  // that case, so we render our own in the presenter's view instead.
+  UIView *_detachedPresenterDimView;
 }
 
 #pragma mark - Initialization
@@ -147,6 +151,9 @@ using namespace facebook::react;
 
   [_snapshotView removeFromSuperview];
   _snapshotView = nil;
+
+  [_detachedPresenterDimView removeFromSuperview];
+  _detachedPresenterDimView = nil;
 
   [TrueSheetModule unregisterViewWithTag:@(self.tag)];
 }
@@ -332,6 +339,9 @@ using namespace facebook::react;
 
   [TrueSheetModule unregisterViewWithTag:@(self.tag)];
 
+  [_detachedPresenterDimView removeFromSuperview];
+  _detachedPresenterDimView = nil;
+
   _lastStateSize = CGSizeZero;
   _didInitiallyPresent = NO;
   _dismissedByNavigation = NO;
@@ -449,6 +459,12 @@ using namespace facebook::react;
   [_screensEventObserver capturePresenterScreenFromView:self];
   [_screensEventObserver startObservingWithState:_state.get()->getData()];
 
+  // If the presenter VC is detached from the window's VC hierarchy, UIKit's
+  // built-in sheet dim renders in the wrong UIWindow. Add our own dim as a
+  // subview of the presenter's view so it lives in the correct window below
+  // the natively-presented sheet.
+  [self addDetachedPresenterDimIfNeededForPresenter:presentingViewController animated:animated];
+
   [presentingViewController presentViewController:_controller
                                          animated:animated
                                        completion:^{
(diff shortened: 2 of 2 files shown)


karimcambridge:
I think I’m having a similar issue! I'm using true-sheet with true-tabs as the footer on my main home screen, then i navigate other screens in the main react-navigation (latest v7) tabs, then when i try to navigate back to home, it says `[Error: Uncaught (in promise, id: 0): "Error: No presenting view controller found"]` I've been fighting for a workaround even using the imperative methods. I think this PR might fix that issue.

lodev09:
I'm not sure about this. I'm trying to avoid functionalities that aren't supported natively, unless it's a real bug/issue. I believe you could also just render a custom backdrop yourself and even integrate reanimated to control the dim.

kanzelm3:
> I'm not sure about this. I'm trying to avoid functionalities that aren't supported natively, unless it's a real bug/issue. I believe you could also just render a custom backdrop yourself and even integrate reanimated to control the dim.

I understand that, but you implemented a workaround for the [sheet presenting below a modal](https://github.com/lodev09/react-native-true-sheet/pull/359)... so not sure why you wouldn't also ensure that dimmed works in that case too. I banged my head against this for hours and realized it was a bug in the library.

lodev09:
> > I'm not sure about this. I'm trying to avoid functionalities that aren't supported natively, unless it's a real bug/issue. I believe you could also just render a custom backdrop yourself and even integrate reanimated to control the dim.
> 
> I understand that, but you implemented a workaround for the [sheet presenting below a modal](https://github.com/lodev09/react-native-true-sheet/pull/359)... so not sure why you wouldn't also ensure that dimmed works in that case too. I banged my head against this for hours and realized it was a bug in the library.

Yeah plenty of hacks on android just to make it consistent with IOS. What I'm avoiding is hacking IOS since things don't usually work correctly. Android is not very opinionated so you can do whatever you want.

On your case, if you could make it work like the normal dim behavior then I'm good with it. Otherwise, I'll just treat it as limitation (which this library has plenty of).

lodev09:
Did you try using custom dim backdrop yourself in JS side?

kanzelm3:
> Did you try using custom dim backdrop yourself in JS side?

I did but ran into timing issues with the fade in/out, will give it another go using the reanimated integration.

PR #664, opened 2026-04-24
add support to render legacy sheet for ios 26+ devices

dhruvpvx:
## Summary

Adds an optional **`preferLegacyStyle`** prop (default `false`) so sheets on **iOS 26+** can opt out of the default liquid glass background and use the same material-style background path as earlier iOS versions. The prop is wired through Fabric (`TrueSheetViewNativeComponent`), `TrueSheet`, and `TrueSheetViewController` (`setupBackground`). **Android** exposes a no-op setter so codegen stays consistent, matching other iOS-only props like `pageSizing`. **`TrueSheetNavigationSheetProps`** includes the new prop so navigator screens can pass it.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- [ ] Build iOS app with New Architecture and confirm no codegen / compile errors after `preferLegacyStyle` is added to the native spec.
- [ ] On **iOS 26+** simulator or device: present a sheet with default props and confirm default appearance (liquid glass path unchanged when `preferLegacyStyle` is omitted or `false`).
- [ ] On **iOS 26+**: set `preferLegacyStyle={true}` on the same sheet and confirm background uses the legacy material path (no liquid glass).
- [ ] On **iOS &lt; 26**: quick smoke test that sheet background behavior is unchanged.
- [ ] **Android**: smoke test that presenting a sheet still works (prop is a no-op).
- [ ] If using **TrueSheet** navigator: pass `preferLegacyStyle` via screen options and confirm it reaches the native view.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web (N/A for this change unless you use web stubs—confirm sheet still loads)
- [x] I updated the documentation (if needed) (`preferLegacyStyle` is documented in `TrueSheet.types.ts` JSDoc)

### DIFF
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
@@ -211,6 +211,11 @@ class TrueSheetViewManager :
     // iOS-specific prop - no-op on Android
   }
 
+  @ReactProp(name = "preferLegacyStyle", defaultBoolean = false)
+  override fun setPreferLegacyStyle(view: TrueSheetView, value: Boolean) {
+    // iOS-specific prop - no-op on Android
+  }
+
   @ReactProp(name = "elevation", defaultDouble = -1.0)
   override fun setElevation(view: TrueSheetView, elevation: Double) {
     view.setSheetElevation(elevation.toFloat())
--- a/ios/TrueSheetView.mm
+++ b/ios/TrueSheetView.mm
@@ -231,6 +231,7 @@ using namespace facebook::react;
   }
 
   _controller.pageSizing = newProps.pageSizing;
+  _controller.preferLegacyStyle = newProps.preferLegacyStyle;
   _controller.dismissible = newProps.dismissible;
   _controller.draggable = newProps.draggable;
   _controller.dimmed = newProps.dimmed;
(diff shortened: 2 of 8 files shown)


lodev09:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

lodev09:
Design question, not a bug: is this prop needed?

On iOS 26.1+, the existing [`setupBackground`](https://github.com/lodev09/react-native-true-sheet/blob/ed47f85ec3b4108f87e037e1530862321cf03089/ios/TrueSheetViewController.mm#L748-L761) already opts out of liquid glass whenever `hasBlur` is true or `backgroundColor` is set — by routing through `sheet.backgroundEffect = UIColorEffect(clearColor)` or `UIColorEffect(backgroundColor)`.

So users can already achieve the same result today:

```tsx
// Equivalent to preferLegacyStyle={true}
<TrueSheet backgroundBlur="systemMaterial" />
// or
<TrueSheet backgroundColor="#1c1c1e" />
```

The only scenario where `preferLegacyStyle=true` changes behavior is when **neither** `backgroundBlur` nor `backgroundColor` is set — in which case the [new outer `else if`](https://github.com/lodev09/react-native-true-sheet/blob/ed47f85ec3b4108f87e037e1530862321cf03089/ios/TrueSheetViewController.mm#L737-L739) forces `SystemMaterial`. So this is effectively sugar for "default to SystemMaterial instead of liquid glass" when no other styling is specified.

If that's the intended value (a cleaner intent-based API vs. hardcoding `backgroundBlur="systemMaterial"`), fine. But worth considering whether documenting the `backgroundBlur` escape hatch instead would avoid introducing a prop whose behavior overlaps with existing ones.

🤖 Generated with [Claude Code](https://claude.ai/code)

lodev09:
I thought this was redundant and unnecessary, claude thinks so too. Closing

PR #665, opened 2026-04-24
feat(ios): expose UIGlassEffect clear variant via new `backgroundGlass` prop

### LINKED ISSUE #663: feat(ios): expose UIGlassEffect clear variant via new `glassStyle` prop
## Summary

Expose iOS 26's `UIGlassEffect.Style.clear` variant via a new `glassStyle` prop so apps can opt into the high-transparency Liquid Glass material for media-rich sheets.

Currently on iOS 26+, when neither `backgroundColor` nor `backgroundBlur` is set, TrueSheet lets the system apply its default Liquid Glass — which is `UIGlassEffectStyleRegular` under the hood (material + luminosity overlay + automatic tint). That's the right default, but there's no way to opt into `.clear`.

## Why it matters

iOS 26's `UIGlassEffect` ships two styles:

| Style | Use case | Visual |
|---|---|---|
| `.regular` | Default. Chrome, generic sheets, nav bars. | Blur + white luminosity veil, strong legibility baseline. |
| `.clear` | Media-rich backgrounds (photos, video, maps). | High transparency, limited adaptivity, meant to be paired with a manual dimming/blur layer. |

Apple's own HIG explicitly calls out `.clear` for "bold content over rich backgrounds". For apps that present a sheet over a custom animated canvas (our case: a football pitch with animated crowd / LED jumbotron, see screenshot below) the regular veil washes out the background. `.clear` + app-side thin blur gives an editorial result that feels native.

Right now the only workarounds are:

1. Set `backgroundColor` → disables Liquid Glass entirely, loses rim refraction
2. `backgroundBlur="system-ultra-thin-material"` → legacy UIBlurEffect path, loses the iOS 26 material
3. Patch the library (what we're currently doing via `patch-package`)

None of these let us keep iOS 26 Liquid Glass rim + clear variant.

## Proposed API

```tsx
<TrueSheet
  detents={[0.6, 0.95]}
  glassStyle="clear"  // ← new, defaults to 'regular'
  cornerRadius={20}
  scrollable
>
  {/* app-side dimming + content */}
</TrueSheet>
```

- New prop `glassStyle?: 'regular' | 'clear'`, default `'regular'`
- Ignored when `backgroundColor` or `backgroundBlur` is set (explicit precedence, matches existing behavior)
- Ignored on pre-iOS 26 runtimes (silently falls back to current path)
- Android / web: no-op

## Reference implementation

I've shipped this in our app via `patch-package` and it works cleanly. Happy to open a PR — here's the gist of the native change:

**`TrueSheetViewController.mm` (`setupBackground`):**

```objc
#if RNTS_IPHONE_OS_VERSION_AVAILABLE(26_1)
  if (@available(iOS 26.1, *)) {
    if (!self.isDesignCompatibilityMode) {
      if (self.backgroundColor) {
        self.sheet.backgroundEffect = [UIColorEffect effectWithColor:self.backgroundColor];
      } else if (hasBlur) {
        self.sheet.backgroundEffect = [UIColorEffect effectWithColor:[UIColor clearColor]];
      } else if (self.glassStyle == facebook::react::TrueSheetViewGlassStyle::Clear) {
        UIGlassEffect *glass = [UIGlassEffect effectWithStyle:UIGlassEffectStyleClear];
        glass.interactive = YES;
        self.sheet.backgroundEffect = glass;
      } else {
        self.sheet.backgroundEffect = nil; // system default = regular
      }
      return;
    }
  }
#endif
```

Plus:
- New prop in `src/fabric/TrueSheetViewNativeComponent.ts` spec: `glassStyle?: WithDefault<'regular' | 'clear', 'regular'>`
- New type `GlassStyle` in `src/TrueSheet.types.ts`
- Forward in `src/TrueSheet.tsx`
- Add to `TrueSheetNavigationSheetProps` Pick list
- `@property` on `TrueSheetViewController.h`, wire in `TrueSheetView.mm`

Full diff: ~220 lines (patch-package file), no behavior change for existing users, purely additive.

## Would a PR be welcome?

If yes, I'll send one with docs + a screenshot in the README's iOS 26 section. Let me know preferred naming (`glassStyle` vs `liquidGlassStyle` vs something else) and whether you'd want `isInteractive` exposed too (I default it to `YES` internally, matching Apple's typical usage for draggable sheets).

Thanks for the library — it's doing real work for us on iOS 26.

michelemarri:
Closes #663

Adds a new `backgroundGlass?: 'regular' | 'clear'` prop on `<TrueSheet>` so apps can opt into iOS 26's `UIGlassEffectStyleClear` variant. Naming mirrors the existing `backgroundColor` / `backgroundBlur` props per your suggestion in the issue thread.

## API

```tsx
<TrueSheet detents={[0.5, 0.9]} backgroundGlass="clear">
  {/* ... */}
</TrueSheet>
```

- **'regular'** (default): existing behavior — system-default Liquid Glass (`.regular` style, with luminosity veil for content legibility)
- **'clear'**: new option → `UIGlassEffect(style: .clear, isInteractive: true)`. High transparency, minimal adaptivity, designed for media-rich backdrops where the app provides its own dimming/blur layer

## Precedence

Follows the existing pattern — `backgroundColor` wins over `backgroundBlur`, which wins over `backgroundGlass`. The new prop only kicks in when neither of the other two is set, on iOS 26.1+, and `UIDesignRequiresCompatibility` is not opted-in.

## Scope

- `src/fabric/TrueSheetViewNativeComponent.ts` — Fabric spec
- `src/TrueSheet.types.ts` — public TS types (new `BackgroundGlass` type + prop)
- `src/TrueSheet.tsx` — destructure + forward to native
- `src/navigation/types.ts` — add to `TrueSheetNavigationSheetProps` Pick list
- `ios/TrueSheetViewController.h` — new `@property`
- `ios/TrueSheetView.mm` — wire `newProps.backgroundGlass → _controller.backgroundGlass`
- `ios/TrueSheetViewController.mm` — use it in `setupBackground`
- `android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt` — no-op `setBackgroundGlass` stub (iOS-only prop)

~44 lines added, zero deletions. Fully additive: no behavior change for existing users, `default 'regular'` preserves current Liquid Glass appearance. Android gets a no-op `setBackgroundGlass` stub so the Fabric prop is a clean cross-platform no-op there.

## Testing

Used via `patch-package` in our app — **x72**, an Italian fantasy-football app for Serie A — on iOS 26 devices, ahead of our App Store launch. The `backgroundGlass="clear"` variant backs a sheet presented over a custom animated pitch canvas, where the system's regular glass veil was washing out the background. No regressions on existing sheets that don't set the prop.

**Note — clear glass is subtle by design.** `UIGlassEffectStyleClear` is near-transparent with no luminosity veil, so over a plain/solid background it can look identical to `.regular` (or to no effect at all). It only reads as different when there's rich/media content *behind* the sheet **and** the app supplies its own dimming/blur layer — the use case Apple's HIG calls out for clear glass. In our app we pair it with a thin `BlurView` + a team-color gradient over the pitch; without such a backdrop the prop is effectively a visual no-op. (I suspect this is behind the "doesn't render" reports — happy to share a side-by-side screenshot.)

## Out of scope (follow-up)

A `backgroundGlassAppearance` / forced-dark option is intentionally left out here to keep this PR minimal and additive — better suited to a follow-up `backgroundGlassOptions` prop, per the thread discussion. Overriding the interface style is orthogonal to the clear/regular choice.

Thanks for the library and the quick turnaround on the issue!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🎉 New features
+
+- **iOS**: Add `backgroundGlass` prop to opt into iOS 26's clear Liquid Glass variant (`UIGlassEffectStyleClear`). ([#665](https://github.com/lodev09/react-native-true-sheet/pull/665) by [@michelemarri](https://github.com/michelemarri))
+
 ### 🐛 Bug fixes
 
 - **Android**: Fixed focused input in sheet causing auto-focus on main screen input after dismiss. ([#649](https://github.com/lodev09/react-native-true-sheet/pull/649) by [@lodev09](https://github.com/lodev09))
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewManager.kt
@@ -191,6 +191,11 @@ class TrueSheetViewManager :
     // iOS-specific prop - no-op on Android
   }
 
+  @ReactProp(name = "backgroundGlass")
+  override fun setBackgroundGlass(view: TrueSheetView, value: String?) {
+    // iOS-specific prop (liquid glass) - no-op on Android
+  }
+
   @ReactProp(name = "blurOptions")
   override fun setBlurOptions(view: TrueSheetView, options: ReadableMap?) {
     // iOS-specific prop - no-op on Android
--- a/ios/TrueSheetView.mm
+++ b/ios/TrueSheetView.mm
@@ -187,6 +187,9 @@ using namespace facebook::react;
   // Blur tint
   _controller.backgroundBlur = newProps.backgroundBlur;
 
+  // Liquid Glass style (iOS 26+)
+  _controller.backgroundGlass = newProps.backgroundGlass;
+
   // Blur options
   const auto &blurOpts = newProps.blurOptions;
   _controller.blurIntensity = blurOpts.intensity >= 0 ? @(blurOpts.intensity) : nil;
--- a/ios/TrueSheetViewController.h
+++ b/ios/TrueSheetViewController.h
@@ -65,6 +65,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign) BOOL dimmed;
 @property (nonatomic, strong, nullable) NSNumber *dimmedDetentIndex;
 @property (nonatomic, assign) facebook::react::TrueSheetViewBackgroundBlur backgroundBlur;
+@property (nonatomic, assign) facebook::react::TrueSheetViewBackgroundGlass backgroundGlass;
 @property (nonatomic, strong, nullable) NSNumber *blurIntensity;
 @property (nonatomic, assign) BOOL blurInteraction;
 @property (nonatomic, assign) BOOL pageSizing;
--- a/ios/TrueSheetViewController.mm
+++ b/ios/TrueSheetViewController.mm
@@ -746,7 +746,16 @@ static BOOL TrueSheetPositionStateEquals(TrueSheetPositionState a, TrueSheetPosi
         self.sheet.backgroundEffect = [UIColorEffect effectWithColor:self.backgroundColor];
       } else if (hasBlur) {
         self.sheet.backgroundEffect = [UIColorEffect effectWithColor:[UIColor clearColor]];
+      } else if (self.backgroundGlass == facebook::react::TrueSheetViewBackgroundGlass::Clear) {
+        // Clear glass variant (iOS 26+). Higher transparency, no luminosity
+        // veil. Designed for media-rich backgrounds; typically paired with
+        // an explicit dimming/blur layer for content legibility.
+        UIGlassEffect *glass = [UIGlassEffect effectWithStyle:UIGlassEffectStyleClear];
+        glass.interactive = YES;
+        self.sheet.backgroundEffect = glass;
       } else {
+        // Default: let the system apply iOS 26's standard Liquid Glass
+        // (UIGlassEffectStyleRegular under the hood).
         self.sheet.backgroundEffect = nil;
       }
       return;
--- a/src/TrueSheet.tsx
+++ b/src/TrueSheet.tsx
@@ -441,6 +441,7 @@ export class TrueSheet extends PureComponent<TrueSheetProps, TrueSheetState> {
       initialDetentAnimated = true,
       dimmedDetentIndex,
       backgroundBlur,
+      backgroundGlass,
       blurOptions,
       cornerRadius,
       maxContentHeight,
@@ -487,6 +488,7 @@ export class TrueSheet extends PureComponent<TrueSheetProps, TrueSheetState> {
         style={styles.sheetView}
         detents={resolvedDetents}
         backgroundBlur={backgroundBlur}
+        backgroundGlass={backgroundGlass}
         blurOptions={blurOptions}
         backgroundColor={backgroundColor}
         cornerRadius={cornerRadius}
--- a/src/TrueSheet.types.ts
+++ b/src/TrueSheet.types.ts
@@ -194,6 +194,16 @@ export type InsetAdjustment =
   | 'never';
 
 /**
+ * iOS 26+ Liquid Glass style.
+ *
+ * - 'regular': Standard glass with luminosity veil for readability (default).
+ * - 'clear': High-transparency glass for media-rich backgrounds.
+ *
+ * @platform ios 26+
+ */
+export type BackgroundGlass = 'regular' | 'clear';
+
+/**
  * Blur style mapped to native values in IOS.
  *
  * @platform ios
@@ -370,6 +380,15 @@ export interface TrueSheetProps extends ViewProps {
    * @platform ios
    */
   backgroundBlur?: BackgroundBlur;
+
+  /**
+   * iOS 26+ Liquid Glass style. Ignored when `backgroundColor` or
+   * `backgroundBlur` is set, or on pre-iOS 26 runtimes.
+   *
+   * @platform ios 26+
+   * @default 'regular'
+   */
+  backgroundGlass?: BackgroundGlass;
 
   /**
    * Options for customizing the blur effect.
--- a/src/fabric/TrueSheetViewNativeComponent.ts
+++ b/src/fabric/TrueSheetViewNativeComponent.ts
@@ -85,6 +85,10 @@ export interface NativeProps extends ViewProps {
     'none'
   >;
 
+  // iOS 26+ Liquid Glass style. Ignored when backgroundColor or
+  // backgroundBlur is set, or on pre-iOS 26 runtimes.
+  backgroundGlass?: WithDefault<'regular' | 'clear', 'regular'>;
+
   anchor?: WithDefault<'left' | 'center' | 'right', 'center'>;
   anchorOffset?: WithDefault<Double, 16>;
   insetAdjustment?: WithDefault<'automatic' | 'never', 'automatic'>;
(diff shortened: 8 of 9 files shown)


lodev09 (changes requested):
Hey @michelemarri, thanks for the PR! Looks good to me.
Have you tried building on android? I think this will break, we need a stub for this on android, I think.

Also if you don't mind updating the CHANGELOG. Thank you!

bigcupcoffee:
Hey I'm also interested in the feature, so I can take this PR over if the OP won't @lodev09 
Won't we also need something like `backgroundGlassAppearance` prop to support specifying `dark` manually? Essentially a way to do `overrideUserInterfaceStyle` on-demand, as from what I see there's no way to force something like dark theme on dark-only screen when your general app theme is light right now?

On Android we're only missing
```kt
@ReactProp(name = "backgroundGlass")
override fun setBackgroundGlass(view: TrueSheetView, tint: String?) {
  // iOS-specific prop - no-op on Android
}
```

lodev09:
@bigcupcoffee feel free to open up separate PR, looks like @michelemarri is MIA lol. 

> Won't we also need something like backgroundGlassAppearance prop to support specifying dark manually?

I'm not sure if this is available on IOS? If it is, we could add separate options prop for this e.g. `backgroundGlassOptions`.

bigcupcoffee:
It's not an explicit property, just a basic `view.overrideUserInterfaceStyle` should suffice although I'm not sure if I have a view to put it on, I'll play around and find out in a bit I guess

bigcupcoffee:
Hm I couldn't get it to render correctly using 26.1.1, either I need to try newer version or `backgroundEffect` doesn't work that way?

UPD: I'll revisit it with newer xCode later

bigcupcoffee:
@lodev09 retried on Xcode 26.3, no dice, I don't think this PR is working 🤔 Consider trying yourself/closing this

michelemarri:
@bigcupcoffee @lodev09 sorry for going quiet — picking this back up.

It *is* working for us. We've been using this exact change via `patch-package` in our app (**x72**, an Italian fantasy-football app for Serie A) across builds on iOS 26 devices, ahead of our App Store launch. The `clear` variant backs the predictions sheet on our match screen.

I think the "no dice" is the effect being **invisible rather than broken**. `UIGlassEffectStyleClear` is deliberately near-transparent with no luminosity veil, so over a plain/solid background it looks basically identical to `.regular` (or to no effect). It only reads as different when (1) there's rich/media content *behind* the sheet, and (2) the app provides its own dimming/blur layer — exactly the use case Apple's HIG calls out for clear glass.

In our case the sheet sits over an animated pitch canvas, and we layer a thin `BlurView` (intensity ~20) + a team-color gradient as `absoluteFill` children inside the sheet. With that backdrop the `regular` → `clear` difference is obvious; without it, it's a visual no-op. The native side is just:

```objc
UIGlassEffect *glass = [UIGlassEffect effectWithStyle:UIGlassEffectStyleClear];
glass.interactive = YES;
self.sheet.backgroundEffect = glass;
```

applied only when neither `backgroundColor` nor `backgroundBlur` is set, on iOS 26+. Happy to drop a side-by-side screenshot if that helps confirm it.

On the open questions:
- **Android** — good catch, this branch doesn't have it yet. I already have the no-op stub locally (matching your snippet exactly), so I'll push it to this branch.
- **`backgroundGlassAppearance` / forcing dark** — I'd keep that out of this PR to stay minimal/additive and do it as a follow-up `backgroundGlassOptions` prop, per @lodev09's suggestion. Overriding the interface style is orthogonal to the clear/regular choice.

@bigcupcoffee if you'd still rather take it over that's totally fine — just flagging it's not dead and the effect is real, it's only subtle by design.

bigcupcoffee:
I was testing it over the map example in the Example app, which is not a solid background so it would've shown?

Instead it looks like `backgroundColor="transparent"`, no border no nothing. Side by side screenshots would be nice, especially without overlaying another blur view

Thank you!

bigcupcoffee:
Also generally where I'm coming from - I'm pretty sure I tried setting regular effect manually in the code as well (instead of just ignoring the path and letting iOS take it over), to confirm that `backgroundEffect = glass` works at all. IIRC it rendered the exact same no-background sheet, which heavily implied to me that the prop is being misused and customization like `tintColor` is basically impossible, correct me if I'm wrong

lodev09:
I tried it myself and yeah it's very subtle, looked like a `backgroundColor=transparent` visually. However, if this works for you, we could still add this since this is supported by IOS. 

@michelemarri if you could update the docs as well for the prop and also some notes about subtle result.

bigcupcoffee:
> we could still add this since this is supported by IOS.

I don't think it is, to me it looks like iOS falls back to no effect due to invalid `backgroundEffect`. Here's me trying to apply "Regular" effect on this glass path, also applying `tintColor`:
image
 
None apply, this is not the correct or documented use of `backgroundEffect` and there's no blur (not even a slight one) applied, it's just `backgroundColor="transparent"`, which the author should probably use for their case

lodev09:
@bigcupcoffee interesting. So technically ios supports the code (no compile error) but not supported visually and falls back to nil state -- which I think a bug on ios lol.

michelemarri:
Followed up on @bigcupcoffee's ask for a clean side-by-side. I built a neutral repro in **this library's own Example app** (`MapScreen`, sheet floating over the map — **no x72 code**) on **Xcode 26.5 / iOS 26.5**, with a 3-way toggle on the sheet so the variants are unambiguous:

| `regular` (default) | `clear` | `clear` + dimming/blur layer |
|:---:|:---:|:---:|
| regular | clear | clear plus blur |

**What each mode actually does:**
- **`regular`** — system Liquid Glass *with* the luminosity veil → the frosted card; the map behind is washed out.
- **`clear`** — the veil is removed → highly translucent. On its own it reads as near-transparent. **@bigcupcoffee you were right**: over bare content `clear` looks like `backgroundColor: "transparent"`. That's not a bug — it's what the clear glass variant *is*.
- **`clear` + an app-supplied dimming/blur layer** — the veil-less-but-frosted look: the media still shows through, content stays legible.

This lines up with Apple's own guidance for the clear variant ([HIG → Materials](https://developer.apple.com/design/human-interface-guidelines/materials)):
> The `clear` variant is highly translucent, which is ideal for prioritizing the visibility of the underlying content…
> Use this variant for components that float above media backgrounds — such as photos and videos — to create a more immersive content experience.
> If the underlying content is bright, consider adding a dark dimming layer of 35% opacity.

So `backgroundGlass="clear"` isn't meant to render a visible frosted card by itself — it's the **opt-out from the mandatory iOS 26 veil**, so the app can supply its own dimming/material the way Apple recommends. It's the only way to get there on a sheet today (the alternatives are `backgroundColor` = solid, or `backgroundBlur` = a different effect). The third screenshot is just `clear` + a plain `BlurView` over the whole sheet; in x72 we use the same idea (a light blur over an animated background).

Net: the prop behaves correctly — the "nothing renders" reports are the clear variant working as designed (veil removed) without an app-supplied dimming layer on top. If you'd like, I'm happy to add a small toggle like this to the Example so the three variants are discoverable in-repo.

bigcupcoffee:
Uhh, you're telling me that this:
image

Is the same effect as this?
image

This is clearly not the clear variant of liquid glass, as we'd actually see it. Also there's no liquid glass interaction ripple/bubble effect, which you've clearly set to `YES` in your PR

The frosted look you're going for is probably achievable with `backgroundBlur` and `blurOptions`, as you're just overlaying a blur layer on top of literal transparency

lodev09:
@bigcupcoffee thanks for the thorough  investigation -- looks like you're right on this one, ios does supports the code but not the functionality, yet.

@michelemarri closing this one, please try out the `transparent` approach instead and let me know how it goes.

Thanks both!

PR #666, opened 2026-04-26
fix(android): emit didFocus after parent translate animation

lodev09:
## Summary

On Android, when a child sheet is dismissed in a stack, the parent's `onDidFocus` was firing immediately at the end of the child's dismiss animation — *before* the parent's translate-up animation began. It now fires after the parent's translate animation completes, so JS sees the focus event when the parent has visually settled back into place.

Refactored the dismiss flow so event orchestration lives in `TrueSheetView.viewControllerDidDismiss` instead of leaking into `resetTranslation`:

- `translateSheet` accepts an optional `onEnd` callback wired through `withEndAction`
- `resetTranslation` stays pure — translate + cascade only
- `StackManager.unregisterSheet` is just a registry op (no return value)
- The actual `parent` reference is passed through the delegate (was `hadParent: Boolean`), so cross-container scenarios target the correct sheet for both translation and the focus event
- Guarded with `isPresented && !isBeingDismissed` so cascade dismiss (parent dismissing while child dismisses on top of it) doesn't fire focus on a dying parent

## Type of Change

- [x] Bug fix

## Test Plan

- [x] Stacked sheet dismiss → parent's `onDidFocus` fires after parent slides back up
- [x] Single sheet dismiss → no `onDidFocus` event (unchanged)
- [x] `dismissAll` cascade → no spurious `onDidFocus` on dying parents

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐛 Bug fixes
 
+- **Android**: Emit `onDidFocus` after the parent's translate-up animation completes when a stacked child sheet is dismissed. ([#666](https://github.com/lodev09/react-native-true-sheet/pull/666) by [@lodev09](https://github.com/lodev09))
 - **Android**: Fixed focused input in sheet causing auto-focus on main screen input after dismiss. ([#649](https://github.com/lodev09/react-native-true-sheet/pull/649) by [@lodev09](https://github.com/lodev09))
 
 ## 3.10.0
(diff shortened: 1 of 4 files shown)


lodev09:
:rocket: _This pull request is included in v3.10.1. See [Release 3.10.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.10.1) for release notes._

PR #685, opened 2026-05-17
Fix Mac Catalyst build issue

theeket:
## Summary

Fix Mac Catalyst build failure with the following error:

```
'backgroundEffect' is unavailable: not available on macCatalyst
```

This can easily be fixed by wrapping with:
```objc
#if !TARGET_OS_MACCATALYST
```

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

After applying this patch, `react-native-true-sheet` works perfectly fine on Mac Catalyst as a native macOS modal, with a few platform-specific behaviors (which is perfectly fine):

* Always animated from top as native modal.
* Not dismissible regardless of the `dismissible` prop (native behavior). It can still be dismissed programmatically via `.dismiss()`.
* Always uses full window height regardless of `detents`. The height can still be adjusted using `maxContentHeight`.

## Screenshots / Videos

Before Fix:

After Fix (Example):

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - **Web**: Fixed detached/form sheet autopresent animation, scrollable content, footer click, and grabber drag interactions. ([#684](https://github.com/lodev09/react-native-true-sheet/pull/684) by [@lodev09](https://github.com/lodev09))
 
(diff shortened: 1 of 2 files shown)


lodev09 (approved):
Thanks @theeket

PR #696, opened 2026-05-28
fix(ios): smooth sheet position jump when keyboard dismisses with focus outside sheet

Pekaz:
## Summary

Fix a visual glitch where the sheet would abruptly jump when the keyboard dismissed after the first responder had already moved out of the sheet (e.g. tapping outside the input area).

Two issues were combined:

1. The early-return checked only `isFirstResponderWithinSheet`, so any keyboard hide that arrived *after* focus had left the sheet was dropped entirely. Delegates never received `keyboardWillHide`, leaving the sheet in a stale "keyboard-visible" state.

2. Relying on `UIKeyboardWillChangeFrameNotification` alone occasionally missed the hide signal in this scenario.

### Changes

- Subscribe to `UIKeyboardWillHideNotification` in addition to `UIKeyboardWillChangeFrameNotification`.
- Track whether the sheet itself was holding the keyboard (`_isKeyboardVisibleForSheet`) so hide events still propagate after the first responder has moved out.
- Add a `smoothKeyboardHide…` helper that absorbs the position offset between the sheet's current frame and its snapped detent target via a `CGAffineTransform` translation, animated with the keyboard's duration and curve.
- Only emit `keyboardWillHide` to delegates when the sheet was actually involved with the keyboard, avoiding spurious layout work for unrelated hide notifications.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- [x] Focus a `TextInput` inside the sheet, then tap outside the input. keyboard dismisses smoothly without the sheet jumping

## Screenshots / Videos

- Before

https://github.com/user-attachments/assets/48efd885-30fe-43e9-94a5-2d69777367ef

- After

https://github.com/user-attachments/assets/4e39dc3b-039c-42ce-a2d6-ed393c177a69

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - **iOS**: Fixed Mac Catalyst build issue. ([#685](https://github.com/lodev09/react-native-true-sheet/pull/685) by [@theeket](https://github.com/theeket))
 - **iOS**: Fixed footer rendering at the bottom of the content view instead of the bottom of the sheet when the footer view is recycled across present cycles. ([#688](https://github.com/lodev09/react-native-true-sheet/pull/688) by [@lucaswickstrom](https://github.com/lucaswickstrom))
+- **iOS**: Fixed sheet position jump when the keyboard dismisses after the first responder has moved out of the sheet. ([#696](https://github.com/lodev09/react-native-true-sheet/pull/696) by [@Pekaz](https://github.com/Pekaz))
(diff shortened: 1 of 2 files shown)


lodev09:
@Pekaz this is a regression issue from RN, see this previous PR (with the patch):
https://github.com/lodev09/react-native-true-sheet/pull/632

Troubleshooting: https://sheet.lodev09.com/troubleshooting#sheet-snaps-without-animation-on-keyboard-dismiss-ios

PR #716, opened 2026-06-19
feat: add footerOptions.keyboardOffset

bigcupcoffee:
## Summary

Add `footerOptions.keyboardOffset` so we can offset the safe area bottom padding we usually apply when keyboard is open.

Scrollable sheets without a footer when raised do not have any extra bottom space you can scroll to

If prop is updated while keyboard is open Android will update the position in real-time, while iOS won't (no layout update or keyboard event happens). I didn't add any extra logic to handle it on iOS, let me know if (and ideally how) you'd want that too

## Type of Change

- [x] New feature

## Test Plan

Simulators

## Screenshots / Videos

Both iOS/Android videos first shows before using it, then after (simple footer with `paddingBottom: bottomInset + GAP`)

| iOS | Android |
| -- | -- |
| ios.mov | android.mp4 |

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🎉 New features
 
-- Added `footerOptions.keyboardOffset` to reduce how far the footer rises when the keyboard opens, so its safe-area padding tucks behind the keyboard instead of leaving a gap. ([#716](https://github.com/lodev09/react-native-true-sheet/pull/716) by [@bigcupcoffee](https://github.com/bigcupcoffee))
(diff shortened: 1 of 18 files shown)


lodev09:
@bigcupcoffee thanks for adding this! looks good.. could you also add some use-case in the example too?

bigcupcoffee:
@lodev09 decided to use it in prompt sheet with the buttons (given the default footer is already used in basic sheet), hope that's okay!

https://github.com/user-attachments/assets/b89e69f6-9519-44f8-a684-6ff99a41ff9a

lodev09:
@bigcupcoffee I looked at it more and realized that providing a positive number as "offset" here is confusing/inconsistent. I flipped the math to be additive instead, so providing a negative value achieves the same effect.

lodev09 (approved):

lodev09:
:rocket: _This pull request is included in v3.11.1. See [Release 3.11.1](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.1) for release notes._

PR #719, opened 2026-06-29
feat: interactive back-swipe (iOS) and back propagation when not dismissible (Android)

lodev09:
## Summary

Interactive back/dismiss handling for the sheet.

**iOS** — Make the system back-swipe (interactive pop) drive the sheet:
- Track the interactive pop via the screen's transition coordinator and slide the sheet's presentation container with the drag, springing back on cancel and tearing down on commit.
- Run a display link over the gesture that reads the container's presentation-layer translate, so `onPositionChange` tracks the sheet through both the drag and the cancel/commit settle.
- Guard against a re-swipe during settle: track a gesture generation so an interrupted settle animation's late completion can't tear down state owned by a newer back-swipe.

**Android** — Let back propagate when the sheet is not dismissible:
- Previously a non-dismissible sheet collapsed to the first detent on back. Now it leaves the back event unhandled so navigation can go back to the previous screen. `dismissible` sheets still dismiss on back as before; `onBackPress` can still consume the event.

## Type of Change

- [x] Bug fix
- [x] New feature

## Test Plan

- iOS: present a sheet inside a navigation stack, back-swipe from the edge — sheet follows the finger, springs back on cancel, dismisses on commit; `onPositionChange` fires throughout.
- Android: present a non-dismissible sheet containing a navigation stack, press back / swipe back — navigates to the previous screen instead of collapsing.

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### 🎉 New features
+
+- **iOS**: The system back-swipe now interactively drags the sheet closed, tracking the finger and emitting `onPositionChange` through the drag and cancel/commit settle. ([#719](https://github.com/lodev09/react-native-true-sheet/pull/719) by [@lodev09](https://github.com/lodev09))
+
+### 🐛 Bug fixes
+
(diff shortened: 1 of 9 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.2. See [Release 3.11.2](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.2) for release notes._

Issue #705, opened 2026-06-05
Handle growing multiline input?


bigcupcoffee:
### Before submitting a new issue

- [x] I tested using the **latest version** of the library.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of React Native.
- [x] I checked for [existing issues](https://github.com/lodev09/react-native-true-sheet/issues) that might answer my question.

### Bug Summary

Vertically growing multiline inputs do not auto-scroll when typing:

https://github.com/user-attachments/assets/54d9a372-94bd-4c0e-b8e9-8174809d308e

### Affected Platforms

- [x] iOS
- [x] Android
- [ ] Web
- [ ] Other

### Library Version

3.11.0-beta (running example app from main)

### Environment Info

```shell
not needed, example app
```

### Steps to Reproduce

I added this to PromptSheet:
```tsx
<TextInput
  placeholder="Growing TextArea..."
  multiline
  style={{
    backgroundColor: 'rgba(0,0,0,0.5)',
    padding: 16,
    borderRadius: 16,
    color: 'white',
    minHeight: 16 * 5,
  }}
  // RN bug: controlled input is required for it to grow
  value={value}
  onChangeText={setValue}
/>
```

### Repro

https://snack.expo.io/sorrydontwannacreatearepo

### Additional Context

Not sure if you're willing to handle it, it's probably possible to wrap it with keyboard-controller but we'd need to disable all of its layout logic

Also side question: on my RN 0.83 app I can't get Android TextInput to work, it always steals the gesture (I have `scrollable`/`ScrollView.nestedScrollEnabled` and it works correctly in example app so I'm at loss for ideas, any clues are welcome

lodev09:
This is overkill imo. The pr that solves the scroll on input #707 should suffice for this, UX wise.

bigcupcoffee:
That PR is unrelated though, in our app we have keyboard-controller all over the place which handles it perfectly, true-sheet handling keyboard but not this case is UX issue for us

On the video you can also see iOS tries to scroll that input, so you have drag outside of it in order to get to the content you're typing, which might be problematic if that input takes most/all visible sheet space (you'd need to dismiss keyboard and try dragging again)

keyboard-controller seems to have a hook that could let us handle it, which I'll try (otherwise patching it is then)

PR #723, opened 2026-07-06
feat: scroll to input's caret and when user types

bigcupcoffee:
## Summary

**Quick disclaimer:** when I was recording the clips, I encountered a crash when grabbing the caret, upon rebuilding with Xcode to monitor it, it vanished, I'm still somewhat concerned of this, worth testing

This is not a lot of code for a decent UX requirement. Whenever user types it scrolls to the caret

Resolves two issues:
- #705 - scrolls to the caret as user keeps typing
- scrolls to the caret with keyboardOffset in mind, instead of textinput bounds, which handles taller inputs a bit better

This does not handle cases of a caret movement via other means (other than typing), because handling that would require:
- iOS: delegate/subclass proxy OR polling/gesture workarounds
- Android: subclassing ReactEditText (`onSelectionChanged`) OR hijacking its AccessibilityDelegate (`TYPE_VIEW_TEXT_SELECTION_CHANGED`)

This is showcased on the videos - moving caret upwards (on iOS) doesn't scroll to it, user has to type first. On Android upwards is handled by the platform, but moving it down also doesn't scroll to it, making the user type

Since you can achieve an okayish UX with a clean and minimal diff, I figured I'll PR this anyway even though #705 was closed

## Type of Change

- [x] New feature

## Test Plan

Simulators

## Screenshots / Videos

| iOS | Android |
|---|---|
| iOS | Android |

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web


bigcupcoffee:
@lodev09 since it's not that much code does it have a chance to merge?

lodev09:
:rocket: _This pull request is included in v3.11.4. See [Release 3.11.4](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.4) for release notes._

PR #729, opened 2026-07-11
fix(ios): compile tvOS stubs to fix Fabric startup crash

### LINKED ISSUE #722: tvOS support / opt-out of iOS autolinking on TV builds
## Summary

We use `@lodev09/react-native-true-sheet` in an Expo app that ships to **iOS** and **tvOS** via react-native-tvos.

Although bottom sheet is not relevant on tvOS, for those builds (`EXPO_TV=1`) the library's iOS native module is still autolinked.

## Environment

- `@lodev09/react-native-true-sheet`: ^3.11.1
- `react-native`: `react-native-tvos@0.86`

## Problem

Trying to exclude it in `react-native.config.js` like this doesn't work here, because for tvOS, Expo/RN autolinking runs with `--platform ios`, and I don't want to exclude it for iOS:

```js
dependencies: {
  '@lodev09/react-native-true-sheet': {
    platforms: {
      ios: null
    },
  },
},
```

## Request

Would you consider a proper way to let tvOS just be ignored from autolinking, or some other way to avoid issues on tvOS?

lodev09:
## Summary

Fixes #722.

tvOS builds (react-native-tvos) crash at launch with `NSInvalidArgumentException: attempt to insert nil object` in `RCTThirdPartyComponentsProvider thirdPartyFabricComponents`. Codegen runs with `--platform ios` on tvOS and registers TrueSheet components, but the podspec was iOS-only so `NSClassFromString(@"TrueSheetView")` returned nil.

- Add `:tvos` platform to the podspec with per-platform sources: iOS compiles as before; tvOS compiles only `common/cpp` (platform-neutral C++) + a new stub file
- `ios/tvos/TrueSheetStubs.mm` defines inert `RCTViewComponentView` subclasses for all 5 registered components and a stub `TrueSheetModule` (`present` rejects with a clear error; dismiss/resize resolve as no-ops)

Sheets remain unsupported on tvOS — this just makes the library safe to autolink there.

## Type of Change

- [x] Bug fix

## Test Plan

- tvOS: build https://github.com/slvvn/tvos-true-sheet-autolinking-repro against this branch — app should launch without the Fabric registry crash
- iOS: build the example app to confirm the per-platform `source_files` split still picks up all sources

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fixed tvOS builds crashing at launch (`RCTThirdPartyComponentsProvider` nil class). The podspec now supports tvOS by compiling inert stub components — sheets remain unsupported on tvOS. ([#729](https://github.com/lodev09/react-native-true-sheet/pull/729) by [@lodev09](https://github.com/lodev09))
+
 ### 🎉 New features
 
 - **iOS**, **Android**: Scrollable sheets now follow the text caret as the user types, scrolling to the caret line instead of the input's full bounds. ([#723](https://github.com/lodev09/react-native-true-sheet/pull/723) by [@bigcupcoffee](https://github.com/bigcupcoffee))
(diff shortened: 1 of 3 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.4. See [Release 3.11.4](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.4) for release notes._

PR #731, opened 2026-07-12
fix(ios): learn detent offset before present transition emits positions

lodev09:
## Summary

`onPositionChange` emitted a slightly fractional `index` (e.g. `~0.07` instead of `0`) during the present transition, then snapped to the exact detent index right after `didPresent` — visible as a small jump on anything animated with `index` (e.g. a footer translated via reanimated).

Cause: the resolver-vs-actual detent offset (bottom safe area etc.) is only learned in `viewDidAppear`, drag end, and detent change — so the interpolation during the present animation ran against uncorrected detent heights, and the first correct emit happened after present.

Fix: learn the offset at the start of the present transition in `setupTransitionTracker`, before the display link emits positions. The presented frame is already final there (the same assumption `presentedFrame` capture relies on), and the `resolverHeight > 0` guard keeps it a no-op if the detent resolver hasn't run (including iOS < 16).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Reproduced in a consumer app with `['peek', <fraction>]` detents and a footer animated off `onPositionChange` index: footer nudged right after present. With this change the index lands exactly at the target detent and the footer stays put. Regression-checked present → drag between detents → dismiss.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: `onPositionChange` no longer emits a slightly fractional `index` during the present transition — the detent offset is now learned before positions are emitted, so the index lands exactly on the target detent instead of snapping after `didPresent`. ([#731](https://github.com/lodev09/react-native-true-sheet/pull/731) by [@lodev09](https://github.com/lodev09))
+
 ## 3.11.5
 
 ### 🎉 New features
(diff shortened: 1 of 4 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.6. See [Release 3.11.6](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.6) for release notes._

PR #734, opened 2026-07-14
fix(android): allow sheet drag from views overlaying a scrollable

lodev09:
## Summary

On Android, the sheet couldn't be dragged from views overlaying the pinned scrollable — e.g. an absolutely-positioned header (or a `TextInput` inside it) over a `FlatList`. Dragging the list itself worked (nested scrolling), but overlay views consume the touch stream natively, and `BottomSheetBehavior` refuses to capture it because the scrollable's bounds still contain the touch point (`touchingScrollingChild`), so the drag did nothing.

`TrueSheetCoordinatorLayout` now resolves the stream's actual touch target on `ACTION_DOWN` via RN's `TouchTargetHelper` (respects `zIndex` and `pointerEvents`) and intercepts the stream manually — same mechanism as the existing can't-scroll workaround — when the target is:

- inside the sheet,
- outside the pinned scrollable (targets in the list keep nested scrolling), and
- has nothing vertically scrollable along its parent chain (multiline inputs and nested scrollables keep their gestures).

The `TextInput` disallow-intercept swallow from #712 now derives from the same resolved target instead of a geometric search that wrongly consulted the content scrollable — which is why it never worked for scrollable sheets. Maps/sliders keep the standard disallow contract, horizontal gestures keep the existing horizontal lock, and taps are unaffected (interception only engages past touch slop).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

`FlatListSheet` in the example app (header with `TextInput` overlaying a `FlatList`):

- Drag down/up from the header input → sheet drags
- Drag from the empty header space above the input → sheet drags
- Drag the list itself → nested scrolling still expands/collapses the sheet
- Tap the input → focuses and opens keyboard
- Horizontal gestures and RNGH handlers unaffected

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **Android**: The sheet can now be dragged from views overlaying a scrollable — e.g. a header (including a `TextInput` in it) positioned over a `FlatList` — previously the drag did nothing.
+
 ## 3.11.6
 
 ### 🐛 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🐛 Bug fixes
 
-- **Android**: The sheet can now be dragged from views overlaying a scrollable — e.g. a header (including a `TextInput` in it) positioned over a `FlatList` — previously the drag did nothing.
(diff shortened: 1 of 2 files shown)


lodev09:

PR #737, opened 2026-07-15
fix(web): emit target detent position in lifecycle events

lodev09:
## Summary

`onWillPresent` on web emitted an inconsistent `position` — it read the drawer's live `getBoundingClientRect().top` at emit time: `0` on first present (drawer not mounted yet, reads as "fully expanded"), `~windowHeight` on re-present (drawer still offscreen). iOS/Android emit the target detent position.

- Extracted `computeDetentGeometry()` — target top-Y and height ratio per detent, mirroring vaul's snap-offset math exactly (effective height, `maxContentHeight` ceiling, `auto`/`peek` resolution). Shared by `computeDetentInfo` and `interpolateFromPosition` (previously duplicated).
- `onWillPresent`/`onDidPresent`/`onDetentChange` now emit the active detent's target position/detent value — consistent with iOS/Android, even before the drawer mounts. `onDetentChange` previously emitted the *old* position with the new index (fired before the animation); `detent` was `0` for `auto`/`peek` detents.
- Drag events (`onDragBegin`/`onDragChange`/`onDragEnd`) keep the live in-flight position (`computeDetentInfo(true)`), matching native drag semantics, with the detent value now correct for `auto`/`peek`.
- Position interpolation now uses `effectiveDetachedOffset`/`effectiveMaxContentHeight` (was raw props — wrong for form sheets) and resolves `auto` via the tracked content height (vaul's `onContentHeightChange` signal) instead of a DOM query, so it works pre-mount.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Verified event payloads in the example app on web: `willPresent`/`didPresent`/`detentChange` now emit the settled detent position/value across first present, re-present, and resize.
- `yarn typecheck`, `yarn lint`, `yarn test` (44 passed).

## Screenshots / Videos

N/A

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [x] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 🐛 Bug fixes
 
 - **Android**: The sheet can now be dragged from views overlaying a scrollable — e.g. a header (including a `TextInput` in it) positioned over a `FlatList` — previously the drag did nothing. ([#734](https://github.com/lodev09/react-native-true-sheet/pull/734) by [@lodev09](https://github.com/lodev09))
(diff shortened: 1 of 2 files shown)


lodev09:

PR #739, opened 2026-07-15
fix(web): scrollable content cut-off and drag/scroll conflicts during sheet drag

lodev09:
## Summary

Fixes three web drag issues around scrollable content:

- **Mid-drag cut-off**: the scrollable fill is sized off `--snap-point-height`, which only updated on detent change — content stayed at the old visible height and got cut off while dragging. The var now tracks the live drag position and transitions in sync with `transform` (registered via `@property`) so the content resizes smoothly through drags and snap animations.
- **Content scrolling along with the drag (mobile)**: below the last detent, vertical touch pans are disabled ahead of time on the scroll container and its descendants (`touch-action` is latched at gesture start and ancestors above a scroller aren't consulted). At the last detent, scrollers in the touched chain are frozen (`overflow: hidden`) once the sheet drag activates, since browsers ignore `preventDefault` after a scroll gesture is latched. `shouldDrag`'s rest threshold is generalized so sheets with a max detent < 1 still scroll content at rest.
- **Delayed drag after scroll bounce (Safari)**: the rubber-band bounce reports negative `scrollTop`, which armed the scroll lock, and the lock re-armed itself on every prevented move — deadening the whole gesture. Bounce now counts as at-top, prevented attempts no longer re-arm, and the drag re-anchors when it engages mid-gesture so the sheet doesn't jump.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Dragged the FlatList/ScrollView sheets between detents on web — content fills the sheet through the whole drag and snap animation
- On mobile browser: dragging the sheet from scrollable content moves only the sheet; content scrolls normally at the last detent
- On Safari: drag engages immediately during/after the scroll bounce
- Verified drags end-to-end in headless Chrome (CDP touch + mouse) and Playwright WebKit
- `yarn jest` (44 tests), `tsc`, `eslint` pass

## Screenshots / Videos

N/A

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [x] I tested on Web


lodev09:

PR #740, opened 2026-07-15
fix(web): measure initial detent geometry

lodev09:
## Summary

Fix initial Web detent event geometry by measuring auto and peek content from the connected drawer DOM before emitting presentation events. This prevents first-presentation lifecycle events from using fallback dimensions, including during autopresent.

The map example now opens at its first detent to exercise the affected path.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Pre-commit TypeScript checks passed.
- Pre-commit lint completed with no errors.
- Verify a Web sheet using auto and peek detents reports the correct initial position during presentation.

## Screenshots / Videos

Not applicable.

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [x] I tested on Web
- [x] I updated the documentation (if needed)
- [x] I added a changelog entry (if needed)

PR #742, opened 2026-07-15
fix(web): measure peek content live for initial detent geometry

lodev09:
## Summary

#740 deferred the first `willPresent` until the drawer is DOM-connected and live-measured the **header/footer** for `'peek'` detent geometry — but the `TrueSheetPeek` content offset still came from React state fed by RN-web `onLayout`, which dispatches a frame after mount. On autopresent (`initialDetentIndex`), a sheet whose peek includes a `TrueSheetPeek` (e.g. footer + peek content, no header) emitted `willPresent` with only the footer height — the peek content read `0`.

This measures the peek content live from the DOM in `computeDetentGeometry`, same as header/footer:

- `TrueSheetPeekContext` gains a `peekRef` the peek registers its element into.
- The rect-delta measurement (`peek bottom − content top`) is extracted into `measurePeekContentHeight()` and shared with the peek's layout handler. Both rects live in the drawer's transformed subtree, so an in-flight translate cancels out of the delta.
- The live branch condition now includes `peekEl`, covering peek-only sheets (no header/footer) that previously fell back to the state-driven `peekHeight`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Reproduced in a consumer app: autopresented sheet with `detents={['peek', 1]}`, a footer, and `TrueSheetPeek` inside scrollable content — `willPresent` emitted `position = height − footerHeight`. With this fix it emits the full peek position (footer + peek content offset). Verified the patched build in-app.
- Example MapScreen (header-only peek) still reports correct geometry.
- `yarn typecheck`, `yarn lint`, `yarn test` (44 tests) pass.

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [x] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **Web**: The `peek` detent's `TrueSheetPeek` content offset is now measured live during initial presentation — previously the first `willPresent` on autopresent missed it (state lags mount by a frame) and reported only the header/footer height. ([#742](https://github.com/lodev09/react-native-true-sheet/pull/742) by [@lodev09](https://github.com/lodev09))
+
 ## 3.11.8
 
 ### 🐛 Bug fixes
--- a/src/TrueSheet.web.tsx
+++ b/src/TrueSheet.web.tsx
@@ -10,12 +10,9 @@ import {
   useRef,
   useState,
 } from 'react';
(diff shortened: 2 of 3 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.9. See [Release 3.11.9](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.9) for release notes._

PR #751, opened 2026-07-20
feat: v4

lodev09:
v4 layout rewrite — the sheet lays out synchronously per detent with Yoga owning all frames


deniz-blue:
if yoga owns all frames why not allow controlled usage? i.e `<TrueSheet index={state} onIndexChange={setState} />`

PR #765, opened 2026-08-09
fix(android): first tap on Pressable swallowed after touching TextInput

### LINKED ISSUE #764: Android: first tap on a `Pressable` is swallowed after a `TextInput` in the same sheet is touched — `TrueSheetViewController` is missing `requestDisallowInterceptTouchEvent`
## Environment

| | |
|---|---|
| `@lodev09/react-native-true-sheet` | 3.11.9 |
| react-native | 0.86.2 (New Architecture / Fabric) |
| Platform | Android only (verified on a Samsung S23, Android 14) |
| iOS | not affected — this is in the Android view controller |

## Summary

Inside a TrueSheet sheet, once a `TextInput` has been **touched**, the next tap on a sibling
`Pressable` does nothing. An identical second tap works. The control is enabled, on screen,
unoccluded, and correctly positioned — `onPressIn` simply never fires for the first tap.

`TrueSheetViewController` implements React Native's `RootView` and drives its own
`JSTouchDispatcher`, but it does not override `requestDisallowInterceptTouchEvent`. Both of
RN's own root views do, and for exactly this reason.

## Reproduction

1. Render a TrueSheet sheet containing a `TextInput` and a `Pressable` as siblings.
2. Tap the `TextInput` (you do not need to type anything).
3. Dismiss the keyboard (Android back, or `Keyboard.dismiss()`).
4. Tap the `Pressable` **once** → nothing happens.
5. Tap it again → it fires normally.

Minimal shape:

```tsx
<TrueSheet ref={sheet}>
  <View>
    <TextInput placeholder="touch me first" />
    <Pressable onPress={() => console.log("PRESSED")}>
      <Text>Submit</Text>
    </Pressable>
  </View>
</TrueSheet>
```

Expected: one tap on `Submit` logs `PRESSED`.
Actual: the first tap logs nothing; the second logs `PRESSED`.

## Cause

`ReactEditText` calls `parent.requestDisallowInterceptTouchEvent(true)` on `ACTION_DOWN` and
then consumes the touch stream itself. In a normal RN tree that call travels up to the root
view, which **deliberately swallows it** so that it keeps being offered `onInterceptTouchEvent`:

- `ReactRootView.java:454-458` — *"Override in order to still receive events to
  onInterceptTouchEvent even when some other views disallow that, but propagate it up the tree
  if possible."*
- `ReactSurfaceView.kt:132-136` — the same override, same comment.

`TrueSheetViewController` (`TrueSheetViewController.kt:107-109`) implements `RootView` and owns
a `JSTouchDispatcher` (`:192`) which it feeds from `onInterceptTouchEvent` (`:1356`) — but it
inherits `ViewGroup`'s default `requestDisallowInterceptTouchEvent`, which **sets the disallow
flag**. So after the `TextInput`'s `ACTION_DOWN`:

- the controller stops being offered `onInterceptTouchEvent`,
- its `jsTouchDispatcher` therefore never sees the gesture's `UP`,
- JS keeps the `TextInput` as the active responder.

The next `ACTION_DOWN` is then dispatched only to the lowest common ancestor of the stale
responder and the new target, so the sibling `Pressable` is never offered the responder and
`onPressIn` never fires. The second tap works because the stale gesture has by then been
terminated.

**Note the LCA truncation is the consequence, not the cause.** That part is ordinary RN
behaviour — if it were the bug, every RN app would eat the first tap after typing. Plain
screens are immune precisely because `ReactSurfaceView` has the override; we measured two
non-sheet screens in the same app firing on the first tap under identical conditions.

## Diagnostic signal (no instrumentation needed)

RN already logs this exact state at `JSTouchDispatcher.kt:98`:

```
Got DOWN touch before receiving UP or CANCEL from last gesture
```

One gotcha for anyone reproducing: it is emitted under the tag `unknown:ReactNative`, so
`adb logcat -s ReactNative` returns **zero lines** and reads exactly like silence. Filter on
the app's pid instead.

## Suggested fix

Four lines, mirroring `ReactSurfaceView`. This is what we currently ship as a patch:

```kotlin
override fun requestDisallowInterceptTouchEvent(disallowIntercept: Boolean) {
  // Mirrors ReactSurfaceView.requestDisallowInterceptTouchEvent: a RootView must
  // keep being offered onInterceptTouchEvent so its jsTouchDispatcher sees the
  // gesture END, but the request is still forwarded up the tree.
  parent?.requestDisallowInterceptTouchEvent(disallowIntercept)
}
```

Placed next to the existing `dispatchTouchEvent` / `onInterceptTouchEvent` overrides in the
`// MARK: - RootView Touch Handling` section.

Deliberately does **not** set the controller's own flag, and still forwards upward — so
`TrueSheetCoordinatorLayout`'s drag handling is unchanged. We verified the sheet still drags,
dismisses and scrolls normally with the patch applied.

Happy to open a PR if you'd like it in that form.

lodev09:
## Summary

Fixes #764.

`TrueSheetViewController` implements RN's `RootView` and feeds its own `JSTouchDispatcher` from `onInterceptTouchEvent`, but inherited `ViewGroup`'s default `requestDisallowInterceptTouchEvent`. When a `TextInput` inside the sheet called `requestDisallowInterceptTouchEvent(true)` on touch, the controller stopped receiving `onInterceptTouchEvent`, so `jsTouchDispatcher` never saw the gesture's UP — JS kept the `TextInput` as the stale responder and swallowed the next tap on any sibling `Pressable`.

Fix mirrors `ReactSurfaceView.requestDisallowInterceptTouchEvent`: don't set the controller's own disallow flag, forward the request up the tree. `TrueSheetCoordinatorLayout` drag handling still gets the signal.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Reproduced in the example app on Android with a sheet containing a `TextInput` and a sibling `Pressable` counting `onPressIn`/`onPress`:

1. Open the sheet, tap the `TextInput` (no typing).
2. Dismiss the keyboard with the BACK key (a "dismiss keyboard" `Pressable` would absorb the swallowed tap and hide the bug).
3. Tap Submit once.

Before: counters stay `0 · 0`; a second identical tap fires. After: first tap fires `1 · 1`. Control (tap Submit without touching the field) fires on first tap in both builds. Sheet drag, dismiss, and inner scroll verified working with the fix.

## Screenshots / Videos

N/A

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐛 Bug fixes
 
+- **Android**: The first tap on a `Pressable` is no longer swallowed after a `TextInput` in the same sheet is touched — the sheet's root view now forwards `requestDisallowInterceptTouchEvent` up the tree (mirroring `ReactSurfaceView`) so the touch dispatcher sees the gesture end and the stale responder is released. ([#765](https://github.com/lodev09/react-native-true-sheet/pull/765) by [@lodev09](https://github.com/lodev09))
 - Keyboard-driven sheet growth no longer corrupts `animatedIndex` — iOS skips detent offset learning while the keyboard has the sheet grown and re-settles once it's away; Android clamps expected detent tops to the available height. Settle emits now learn at the final frame and bypass the dedupe, so presenting and settling land exactly on the detent index. ([#762](https://github.com/lodev09/react-native-true-sheet/pull/762) by [@lodev09](https://github.com/lodev09))
 
 ## 3.11.9
--- a/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
+++ b/android/src/main/java/com/lodev09/truesheet/TrueSheetViewController.kt
@@ -1347,6 +1347,12 @@ class TrueSheetViewController(private val reactContext: ThemedReactContext) :
     return super.onInterceptTouchEvent(event)
   }
 
+  override fun requestDisallowInterceptTouchEvent(disallowIntercept: Boolean) {
+    // Mirrors ReactSurfaceView: keep receiving onInterceptTouchEvent so
+    // jsTouchDispatcher sees the gesture end, but forward the request up the tree.
+    parent?.requestDisallowInterceptTouchEvent(disallowIntercept)
+  }
+
   override fun onTouchEvent(event: MotionEvent): Boolean {
     eventDispatcher?.let {
       jsTouchDispatcher.handleTouchEvent(event, it, reactContext)


lodev09:

PR #777, opened 2026-08-12
fix: restore sheet scrolling, and stop touch leak on Android

lodev09:
## Summary

Fixes #774, and makes the #767 touch-leak fix actually work on Android.

### Scroll regression (#774)

With `scrollable`, a `ScrollView` whose items are plain `View`s could no longer be scrolled — only items wrapped in `Pressable` responded.

#767 put `onStartShouldSetResponder` on the sheet **container**, making an ancestor of the `ScrollView` the JS responder. That blocks scrolling:

- **iOS** — `RCTScrollViewComponentView._shouldDisableScrollInteraction` walks ancestors for `isJSResponder`; when found, `touchesShouldCancelInContentView:` returns `NO` and the pan never takes over.
- **Android** — `JSResponderHandler.onInterceptTouchEvent` returns `view.id == currentJSResponder`, so the responder view intercepts the gesture.

Touching a `Pressable` worked because it won the responder first (bubble runs deepest-first), so the container never claimed it.

The handlers move to the **host view**. The container is reparented into the native sheet on mount, so the host view is never a native ancestor of the sheet content — the responder negotiation still ends at the sheet boundary, but native gestures are untouched.

### Touch leak on Android (#763, never fixed there)

Tapping the sheet still fired the parent `Pressable`'s `onPress`. Logging the repro showed the sheet claiming the responder correctly, and *then* a second `pointerDown`/press arriving on the parent.

Cause: an event that no React child consumes reaches both `onInterceptTouchEvent` and `onTouchEvent`, and the sheet's roots dispatched to `jsTouchDispatcher` from both — two JS touch streams for one gesture. Such touches resolve to the root's own tag (the host view), so in the second stream React sees that node already holding the responder and, per the lowest-common-ancestor rule, asks only its **ancestors** — handing the touch to the `Pressable` outside the sheet. It also produced the `Got DOWN touch before receiving UP or CANCEL from last gesture` warnings.

`TouchEventDeduper` now guards both nested roots (`TrueSheetViewController`, `TrueSheetFooterView`) so each `MotionEvent` reaches JS once. RN's own `ReactRootView` dispatches from both callbacks too, but a top-level root has no React ancestors to leak to.

The two fixes depend on each other: taps that hit no React child resolve to the host view's tag, so a container-level claim never runs for them.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

1. `scrollable` sheet with plain `View` items — scrolling works
[...]

PR #767, opened 2026-08-09
fix: stop touches inside sheet from leaking to outside touchables

lodev09:
## Summary

Fixes #763. When a `TrueSheet` is rendered inside a `Pressable` (or any touchable), tapping the presented sheet surface fired the parent's `onPress`.

Native views are reparented into the sheet, but RN touch events bubble through the React tree — so touches that nothing inside the sheet claimed kept bubbling up to touchables outside it. The container now:

- Claims unclaimed touches via `onStartShouldSetResponder`, so the responder negotiation never reaches ancestors outside the sheet. Deeper touchables inside the sheet still win (bubble runs deepest-first), and native gestures (sheet drag, scroll) are unaffected since the responder grant doesn't block native handlers.
- Stops raw `onTouch*` events from propagating past the sheet boundary.

Also adds a repro to the example's `ScrollViewSheet`: each item is a `Pressable` with a nested sheet.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

1. Open the ScrollView sheet in the example app
2. Tap an item's text to present the nested sheet
3. Tap the nested sheet surface — the parent item's `onPress` should NOT log
4. Verify buttons, scrolling, refresh control, and sheet dragging inside the sheet still work

## Screenshots / Videos

N/A

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [x] I tested on Web
- [ ] I updated the documentation (if needed)
- [x] I added a changelog entry (if needed)

PR #786, opened 2026-08-18
fix: stop responder negotiation at sheet boundary instead of claiming it

### LINKED ISSUE #782: onPress dead inside a sheet screen since 3.11.10 — reproducible in your own example/expo

### Bug Summary

On iOS, when a sheet screen's content is wrapped in a nested native-stack, every `Pressable` inside that sheet stops firing `onPress` — the pressed state still shows while the button is held, but the callback never runs. A flat `<View>` tree in the same sheet works fine. This started in 3.11.10; 3.11.9 is unaffected.

https://github.com/user-attachments/assets/366383e2-6a12-42ef-a65b-47165240ac81

### Affected Platforms

- [x] iOS
- [x] Android
- [ ] Web
- [ ] Other

### Library Version

3.11.11

### Environment Info

```shell
expo: ^56.0.5 => 56.0.5
expo-router: ~56.2.7 => 56.2.7
react: 19.2.3 => 19.2.3
react-native: 0.85.3 => 0.85.3
@lodev09/react-native-true-sheet: 3.11.11 (workspace)
@react-navigation/native: 7.1.26
@react-navigation/native-stack: 7.14.10 (via expo-router)
react-native-screens: 4.25.2
react-native-gesture-handler: 2.31.2
react-native-reanimated: 4.3.1
react-native-safe-area-context: 5.7.0
```

### Steps to Reproduce

(fork of this one at v3.11.11, single commit on `main`): https://github.com/luckykenlin/react-native-true-sheet

1. In the app, tap **Sheet Navigator**, then **Open Settings Sheet**.

2. Tap any button in that sheet ("Resize to 100%", "Open Profile", "pop()").

3. The button shows its pressed state while held, but `onPress` never fires — the press is visibly registered and then silently dropped.

### Repro

https://github.com/luckykenlin/react-native-true-sheet

### Additional Context

_No response_

lodev09:
## Summary

Fixes #782 — since 3.11.10, `onPress` inside a sheet silently dies when the sheet content contains a nested native-stack (react-native-screens) screen.

**Root cause:** a nested screen inside the sheet isn't mounted under an `RCTRootComponentView`, so react-native-screens attaches its own `RCTSurfaceTouchHandler` to it — one tap produces two JS touch streams. The first grants the responder to the `Pressable`; the second re-runs React's responder negotiation, which skips the current responder and asks its ancestors. The host view's unconditional `onStartShouldSetResponder={() => true}` (added in #767, moved in #777) claimed it, terminating the `Pressable` — pressed state showed, `onPress` never fired. A second finger on the sheet triggered the same steal.

**Fix:** don't claim — stop. `onStartShouldSetResponder`/`onMoveShouldSetResponder` on the host now call `stopPropagation()` and return `false`. React's negotiation loop checks `isPropagationStopped()` between listeners, so:

- In-sheet views are asked **before** the host → touchables/scrollables work normally
- Outside views are asked **after** → the #767 touch leak stays fixed
- The host never becomes responder → nothing to steal, no `setJSResponder` side effects (the scroll-blocking that #777 worked around can't occur by construction)
- The move-negotiation stop also closes the equivalent leak to outside `PanResponder`s

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Reproduced #782 in both example apps (settings sheet content wrapped in a nested native-stack, `detents: [1]` + `scrollable: true`) — buttons pressed but never fired
- With this fix (iOS): nested-stack sheet buttons fire, flat sheets unaffected, scrolling works, taps on the sheet don't leak to touchables behind it
- `yarn typecheck`, `yarn lint`, `yarn test` (44 tests) pass

## Screenshots / Videos

See #782

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- Touchables inside a sheet fire their `onPress` again when the sheet content contains a nested native-stack screen (or a second finger touches the sheet mid-press) — since [#767](https://github.com/lodev09/react-native-true-sheet/pull/767), the sheet claimed the responder whenever the negotiation re-ran mid-gesture, stealing it from the pressed touchable. The sheet now stops the negotiation at its boundary without claiming, which also keeps the touch-leak fix without any `setJSResponder` side effects. ([#786](https://github.com/lodev09/react-native-true-sheet/pull/786) by [@lodev09](https://github.com/lodev09))
+
 ## 3.11.11
 
 ### 🐛 Bug fixes
--- a/src/TrueSheet.tsx
+++ b/src/TrueSheet.tsx
@@ -58,12 +58,21 @@ if (!TrueSheetModule) {
 
 type NativeRef = ComponentRef<typeof TrueSheetViewNativeComponent>;
 
-// Claim touches that no descendant claimed so the responder negotiation
-// doesn't bubble up to touchables outside the sheet
-const absorbUnclaimedTouches = () => true;
-
 // Stop raw touch events from bubbling past the sheet to outside ancestors
 const stopTouchPropagation = (event: GestureResponderEvent) => event.stopPropagation();
+
+// Stop the responder negotiation at the sheet boundary without claiming the
+// responder. Views inside the sheet are asked before this one, so in-sheet
+// touchables work normally; views outside are asked after, so stopping here
+// keeps unclaimed touches from leaking to touchables behind the sheet.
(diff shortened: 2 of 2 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.12. See [Release 3.11.12](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.12) for release notes._

PR #792, opened 2026-08-21
feat: Add lazy prop to control mounting content before presentation

maxlapides:
## Summary

Add a JS-only `lazy` prop (default `true`). Setting `lazy={false}` mounts sheet children while the sheet remains unpresented. Keep the existing lazy behavior by default, begin rendering when `lazy` changes to `false`, preserve mounted content when it changes back to `true`, and keep non-lazy content mounted across dismissals. Do not forward the prop to the native component.

Without this option, a sheet using `detents={['auto']}` cannot mount its content before `present()`. Content that starts with a Suspense fallback, asynchronous loading state, or skeleton can therefore present at the placeholder height and visibly resize when the final content arrives.

Use `lazy={false}` to mount content early, wait for the app's readiness signal such as Suspense resolution, loaded data, or settled `onLayout`, and then call `present()`. This lets the auto detent measure settled React Native content before the presentation animation begins.

Document the new prop and its measurement limitation, add its public type, and cover the mounting and presentation lifecycle in unit tests.

Content that requires window attachment to measure, such as SwiftUI-hosted views rendered through `expo-ui`, still measures only during presentation. React Native/Yoga content was verified to lay out fully while unpresented in a production React Native app on an iOS 26 simulator.

> [!NOTE]
> Originally proposed as a `prewarm` prop; renamed to `lazy={false}` per maintainer feedback.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- [x] `yarn lint`
- [x] `yarn typecheck`
- [x] `yarn test` — 69 tests pass
- [x] Verify the default lazy behavior remains unchanged
- [x] Verify `lazy={false}` renders children while unpresented and when set after mount
- [x] Verify re-enabling `lazy` and dismissing a non-lazy sheet preserve mounted children
- [x] Verify `present()` on a non-lazy sheet skips the additional mount wait
- [x] Verify `lazy` is not forwarded to the native component
- [x] Verify React Native/Yoga content lays out while unpresented on an iOS 26 simulator

## Screenshots / Videos

N/A — this changes when hidden sheet content mounts and does not introduce a visible UI change.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🎉 New features
 
+- Added a JS-only `prewarm` prop to mount sheet content before presentation, allowing `auto` detents to measure settled content before the sheet opens. ([#792](https://github.com/lodev09/react-native-true-sheet/pull/792) by [@maxlapides](https://github.com/maxlapides))
 - **iOS**: Smoother `onPositionChange`, and `onWillDismiss` now fires only when a drag-to-dismiss is committed. ([#744](https://github.com/lodev09/react-native-true-sheet/pull/744), [#756](https://github.com/lodev09/react-native-true-sheet/pull/756) by [@lodev09](https://github.com/lodev09))
 - The `auto` detent now works with plugged scrollables — the sheet sizes to the scrollable's content height. ([#743](https://github.com/lodev09/react-native-true-sheet/pull/743) by [@lodev09](https://github.com/lodev09))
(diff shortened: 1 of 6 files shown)


lodev09:
@maxlapides can we just name this to `lazy={false}` instead? I think that's more clear

maxlapides:
Yes, good idea @lodev09! I renamed it to `lazy` :)

PR #793, opened 2026-08-22
fix(ios): Retry blocked auto-present after recycled controller sessions

### LINKED ISSUE #789: Fast dismiss then re-present strands a late-mounted Fabric subtree until the next React commit (views configured but never inserted)
**Version:** 3.11.11 · RN 0.86.2 (Fabric/new arch) · Expo SDK 57 · iOS 26 physical device · auto-present via `initialDetentIndex={0}` (the app never calls `present()`)

## Reproduction

Complete, minimal repro app (README has device run steps and a demo script): **https://github.com/maxlapides/true-sheet-mount-stall-repro**

The sheet's six rows each mount in two React commits (row + text first, then an `onLayout`-gated `react-native-svg` background — the pattern of any measured decoration). The app mounts a fresh `<TrueSheet initialDetentIndex={0}>` per presentation and unmounts it on `onDidDismiss`. An **Auto fast-cycle** toggle dismisses and immediately remounts continuously; a **Poke (setState)** button triggers a trivial unrelated commit.

- **Expected:** every open shows six labels on dark SVG fills.
- **Actual:** intermittently (physical device, usually within a dozen fast cycles) a sheet opens with the text visible but some or all SVG fills missing. Tapping **Poke** — any React commit — instantly restores all of them. Simulators rarely hit the timing.

## Evidence (from unified-log instrumentation on the RNSVG views in our production app, where we first hit this)

On a broken reopen, all six `RNSVGSvgView`s received their new props while detached (`window == nil`, Fabric recycle pool) and were then **never inserted into the view hierarchy** — no `didMoveToWindow`, no `drawRect`, for 15 minutes while the user stared at the half-rendered sheet. At the next React commit (a tap), all six received their first `didMoveToWindow windowNil=NO` and drew within 2ms, correct props and all — consistent with Fabric's `MountingCoordinator` picking up the missing insertions in the next base-to-latest diff. So the views are configured and ready; the insertion into the hierarchy is what goes missing, which is why redraw-oriented fixes can't touch this.

## Where we believe the race lives (from reading 3.11.11 sources)

- The Fabric container is manually reparented into `_controller.view` (`TrueSheetView.mm` mount path), and native dismissal does not clean that container — teardown completes asynchronously from JS after `onDidDismiss` (`shouldRenderNativeView = false`). A re-present that overlaps this two-owner teardown mounts the new subtree into a container that is mid-teardown.
- `prepareForRecycle` resets four scalar fields but does not detach `_containerView`, reset the pending mount/props/layout flags, or reset `_controller` — a recycled view can inherit a dismissing controller.
- Auto-presentation is attempted only from `didMoveToWindow`; if the retained controller is still `isBeingDismissed`, the method restores the animation flag and schedules no retry.
- #590's `isPresenting` guard covers the imperative `present()` path but not the `initialDetentIndex` auto-present path.

Also worth noting for consumers serializing on the dismiss promise: `dismissAnimated` completes successfully without dismissing when `isBeingDismissed || !isPresented` (so no `onDidDismiss` follows a "successful" dismiss call), and `dismissByRef` rejects `SHEET_NOT_FOUND` after `shouldRenderNativeView` flips.

## Workaround that fixed it for us

Serializing dismiss/re-present at the app layer: queue the latest present until the matching `onDidDismiss` (or the dismiss promise) settles, transition on `onWillDismiss` so gesture dismissals are covered, and mount a fresh sheet instance for the queued presentation rather than reusing the dismissing one. Verified on device: 17/17 consecutive fast reopen cycles attached and drew after the change, vs. reproducing within ~a dozen cycles before.

Happy to provide the full instrumentation traces or test any candidate fix on the device where this reproduces.

maxlapides:
## Summary

A fast dismiss/re-present cycle can recycle the Fabric host while its retained controller is still dismissing. The replacement mount then sees an in-flight controller transition, skips `initialDetentIndex` auto-presentation, and can remain stranded because nothing retries it. Retry initial presentation after attachment, prop updates, or the previous controller session ends, while keeping the previous session's teardown and lifecycle events scoped to its original mount.

Port the `presentInitialIfNeeded` retry shape from #751 to the v3 line for v4 parity. The `prepareForRecycle` hygiene and controller-session ownership handling go beyond what v4 does and are not covered by #751.

Fixes #789

Reproduction: [maxlapides/true-sheet-mount-stall-repro](https://github.com/maxlapides/true-sheet-mount-stall-repro)

## Commits

- `fix(ios): retry blocked auto-present and end recycled controller sessions` adds the retry path, resets recycled mount state, and prevents a retained controller session from emitting lifecycle events through its replacement host.
- `fix(ios): unregister the view tag actually registered with TrueSheetModule` preserves the registered React tag before recycle so the module removes the correct strong-registry entry. This is a separable commit that maintainers can drop, although the recycle path benefits from it.

**Behavior notes:** In a pathological fast cycle where a sheet finishes presenting after its host was already recycled, the previous sheet now snaps closed (`animated:NO`) before the replacement animates in; UIKit cannot cancel an in-flight presentation, and this replaces silently stranding the new sheet.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Verification

- `yarn typecheck`, `yarn lint`, and `yarn test --watchman=false` (44 Jest tests) pass.
- The linked repro app was extended with two auto-cycled shapes: Mode B performs a key-swap remount 30–330 ms into the dismissal animation; Mode C keeps one sheet mounted, swaps its keyed content mid-dismissal, then remounts after dismissal completes to mirror a production sheet-host.
- Red/green testing used Fabric debug builds on a physical iPhone 11 running iOS 26 with React Native 0.86.2. Native `RNSVGSvgView.didMoveToWindow` attach probes were captured through the unified log.
- Unpatched 3.11.12: Mode B wedged 158 of 316 cycles before presentation because auto-present was blocked by `isBeingDismissed` and never retried; the stale dismissal's `onDidDismiss` also emitted into the new mount. Mode C wedged on all 134 of 134 mid-dismissal swaps. The failure reproduced deterministically across the full 30–330 ms sweep.
- Patched: 0 wedges across 325 Mode B cycles and 294 Mode C cycles. Every cycle that presented attached all six second-commit SVG views (6/6), verified from the unified log.
- Limitation: the “presented but SVG fills missing” variant from the original issue report was captured in a production app and did not reproduce in the standalone harness on this hardware with either build. The patched code eliminates every failure this harness can produce.

## Screenshots / Videos

Not applicable; this changes lifecycle handling without changing the sheet's visual design.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: Fast dismiss/re-present cycles no longer strand an `initialDetentIndex` auto-present when the host is recycled while its controller is still dismissing — blocked attempts now retry, and teardown and lifecycle events from the recycled controller session stay scoped to its original mount. ([#793](https://github.com/lodev09/react-native-true-sheet/pull/793) by [@maxlapides](https://github.com/maxlapides))
+
 ## 3.11.12
 
 ### 🐛 Bug fixes
--- a/ios/TrueSheetView.mm
+++ b/ios/TrueSheetView.mm
@@ -41,6 +41,8 @@ using namespace facebook::react;
 @interface TrueSheetView () <TrueSheetViewControllerDelegate,
   TrueSheetContainerViewDelegate,
   RNScreensEventObserverDelegate>
+- (facebook::react::SharedViewEventEmitter)activeEventEmitter;
+- (void)clearStaleControllerSessionOwnershipIfNeeded;
 @end
 
 @implementation TrueSheetView {
@@ -51,6 +53,8 @@ using namespace facebook::react;
   UIView *_snapshotView;
   CGSize _lastStateSize;
   NSInteger _initialDetentIndex;
+  NSUInteger _initialPresentGeneration;
+  NSUInteger _mountGeneration;
   TrueSheetViewInsetAdjustment _insetAdjustment;
   BOOL _scrollable;
   ScrollableOptions *_scrollableOptions;
@@ -58,6 +62,7 @@ using namespace facebook::react;
   BOOL _isSheetUpdatePending;
   BOOL _pendingLayoutUpdate;
   BOOL _didInitiallyPresent;
+  BOOL _controllerSessionOwnedByPreviousMount;
   BOOL _dismissedByNavigation;
   BOOL _pendingNavigationRepresent;
   BOOL _pendingMountEvent;
@@ -85,9 +90,12 @@ using namespace facebook::react;
     _snapshotView = nil;
     _lastStateSize = CGSizeZero;
     _initialDetentIndex = -1;
+    _initialPresentGeneration = 0;
+    _mountGeneration = 0;
     _initialDetentAnimated = YES;
     _scrollable = NO;
     _isSheetUpdatePending = NO;
+    _controllerSessionOwnedByPreviousMount = NO;
 
     _screensEventObserver = [[RNScreensEventObserver alloc] init];
     _screensEventObserver.delegate = self;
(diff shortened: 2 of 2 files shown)


maxlapides:
Upon further testing, the original bug is still reproducible with this patch. I'll keep testing, and I'll let you know once it's ready for review.

maxlapides:
**TL;DR: This PR is now ready for review, and we recommend merging. It fixes three concrete, device-verified TrueSheet defects; the one remaining #789 symptom turned out to be an upstream React Native Fabric bug that no sheet-library change can fix. RN issue to follow.**

## What this PR fixes

1. **Bug:** Fast dismiss → re-present can leave a permanently dead sheet: auto-present via `initialDetentIndex` runs once, is blocked by `isBeingDismissed`, and never retries. **Fix:** Retrying `presentInitialIfNeeded` from `didMoveToWindow`, `finalizeUpdates`, and dismissal completion. **Verification:** On an iPhone 11 running iOS 26 with RN 0.86.2 Fabric, unpatched runs wedged 158/316 mid-dismissal remounts and 134/134 mid-dismissal content swaps; patched runs had 0 wedges in 619 cycles.

2. **Bug:** A previous mount’s dismissal can emit `onWillDismiss`/`onDidDismiss` into the replacement sheet, so apps that remount on `onDidDismiss` tear down the fresh sheet. **Fix:** Controller-session ownership scoping for lifecycle emission. **Verification:** Native lifecycle probes in the patched device run kept dismissal events scoped to the owning controller session while the 619-cycle run completed with 0 wedges.

3. **Bug:** Recycled hosts can inherit the previous mount’s container and pending state. **Fix:** `prepareForRecycle` hygiene. **Verification:** The patched iPhone 11/iOS 26/RN 0.86.2 Fabric run completed 619 cycles with 0 wedges and no recycled-host carryover observed.

These changes port the maintainer’s own v4 (#751) shapes to v3, so the review should map cleanly.

## What this PR does not fix (and why)

The #789 headline symptom— the sheet presents, but a second-commit `react-native-svg` subtree is never inserted until an unrelated later commit— is an upstream React Native Fabric defect. A Delete mutation for a still-mounted view (in our app, a text view inside a session-replay privacy wrapper) throws `'Attempt to recycle a mounted view'` mid-`RCTPerformMountInstructions`; with no exception isolation, the rest of the transaction, including the SVG Create/Insert mutations, is dropped while `MountingCoordinator` has already consumed the revision. In release builds the assert compiles out and the transaction is silently half-applied. No TrueSheet-side change can fix this; TrueSheet’s remount churn merely exposes it. We’ll file the RN issue with the captured evidence and link it from #789.

## v4

We know v4 (#751) is in beta, but this PR still matters on both timelines: v3 is the currently released line (3.11.x patches are still shipping), and several of these fixes address defects also present on the v4 branch. The deferred initial present is not generation-scoped there, dismissal events lack session-ownership scoping, and the module registry unregisters `self.tag` after React Native has already zeroed it, so recycled hosts stay strongly retained.

We’re happy to forward-port fixes 2 and 3, plus the registered-tag fix, to the v4 branch as a follow-up PR if that helps.

Please give the v3 port a final review; it is ready to merge, and we welcome any last comments.

maxlapides:
@lodev09 Okay I finished my investigation, above :) This doesn't exactly fix my original bug, but ended up writing this patch in the process of digging into the issue.

maxlapides:
Since v4 (#751) has merged to main, we retargeted this PR’s base to the new v3 maintenance branch. It was built on the 3.11.12 tip, so it now merges cleanly with no code changes; the offer to forward-port the v4-relevant fixes stands.

maxlapides:
The v4 forward-port offered above is now up as #800, targeting `main`. It carries this PR's lifecycle/recycle fixes and registered-tag cleanup onto the v4 line, plus the v4-specific hardening found during review.

lodev09:
Thanks for the deep dive and the repro. The v4 side of this landed as a narrower fix in #818 (see the note on #800 for what reproduced on `main` and what didn't).

v3 is in maintenance now, so I'm closing this rather than carrying the larger lifecycle rework there. If the mid-dismissal remount case bites on 3.x, a focused port of #818's `prepareForRecycle` detach is the shape I'd take.

PR #800, opened 2026-08-24
fix(ios): retry blocked auto-present after recycled controller sessions

maxlapides:
## Summary

This forward-ports the iOS lifecycle fixes from #793 onto `main` (the v4 line), as offered in that PR's discussion.

A fast dismiss/re-present cycle can recycle the Fabric host while its retained controller is still finishing the previous presentation session. This fixes four concrete failure families:

- **Blocked auto-present:** a replacement mount can find the controller dismissing, skip `initialDetentIndex` auto-presentation, and remain permanently dead because nothing retries it. Initial presentation now retries after attachment, prop updates, or dismissal completion, with deferred work scoped to the current mount and captured presentation intent.
- **Previous-session event leakage:** presentation, dismissal, focus, drag, detent, and position callbacks from the previous mount can otherwise reach the replacement mount. Controller-session ownership now scopes those events and teardown to the mount that started the session.
- **Recycled-host state carryover:** the replacement can inherit the previous mount's container, pending props/detents/layout work, navigation state, or dismissal snapshot. Recycle now detaches the old container, resets per-mount transient state, preserves the snapshot only through the dismissal that needs it, and promotes the replacement's pending detents before its retry presentation.
- **Registered-tag leak:** React Native zeroes `view.tag` before `prepareForRecycle`, so unregistering `self.tag` removed key `0` while `TrueSheetModule`'s strong registry retained every recycled host. The view now stores the tag it registered and unregisters that exact value.

The v4 port also adds review-driven hardening for v4-specific asynchronous paths: the interactive-navigation dismissal completion is mount-scoped; pending detents are applied before retry presentation and active-index clamping; and delayed detent/position emissions are presentation-session-scoped, including the 0.1-second transition-end settle.

This is the v4 counterpart to #793 and fixes the TrueSheet auto-present/lifecycle half of #789. It does **not** claim to fix the remaining #789 symptom where a presented sheet's second-commit subtree is not inserted until a later unrelated commit; as documented in #793, that is an upstream React Native Fabric transaction issue.

The `CHANGELOG.md` entry will be pushed immediately after this PR number exists.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- The underlying lifecycle fixes were device red/green verified on the v3 line in #793 using a physical iPhone 11 running iOS 26 with React Native 0.86.2 Fabric:
  - Unpatched: 158/316 mid-dismissal remounts wedged; 134/134 mid-dismissal content swaps wedged.
  - Patched: 0 wedges across 325 remount cycles and 294 content-swap cycles (619 total).
- This v4 forward-port was verified by static lifecycle analysis and two independent, multi-round code reviews covering host recycle ordering, controller-session ownership, registry lifetime, and delayed callbacks.
- `yarn typecheck` passes.
- `yarn lint` passes.
- `yarn test --watchman=false` passes (65 tests).
- No device soak was run against the v4 branch.

## Screenshots / Videos

Not applicable; this changes lifecycle handling without changing the sheet's visual design.

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web
- [ ] I added a changelog entry (will be pushed immediately after this PR number exists)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### 🐛 Bug fixes
 
+- **iOS**: Fast dismiss/re-present cycles no longer strand an auto-presented sheet or leak the previous mount's lifecycle and detent state into its replacement. Recycled sheet hosts are also released from the native registry correctly. ([#800](https://github.com/lodev09/react-native-true-sheet/pull/800) by [@maxlapides](https://github.com/maxlapides))
 - **Android**: A stacked parent sheet now follows its child frame-by-frame when the child resizes, and no longer gets stuck at stale detents after a half-expanded settle. ([#797](https://github.com/lodev09/react-native-true-sheet/pull/797) by [@lodev09](https://github.com/lodev09))
 - **Android**: The scrollable content now tracks the keyboard frame-by-frame — the focused input rides the keyboard up smoothly, and the scroll position eases back on dismiss instead of snapping. ([#798](https://github.com/lodev09/react-native-true-sheet/pull/798) by [@lodev09](https://github.com/lodev09))
 - The keyboard-driven scroll inset now accounts for a floating footer's height. ([#752](https://github.com/lodev09/react-native-true-sheet/pull/752) by [@lodev09](https://github.com/lodev09))
--- a/ios/TrueSheetModule.mm
+++ b/ios/TrueSheetModule.mm
@@ -152,7 +152,7 @@ RCT_EXPORT_MODULE(TrueSheetModule)
       TrueSheetView *rootSheet = nil;
 
       for (TrueSheetView *view in viewRegistry.allValues) {
-        if (!view.viewController.isPresented) {
+        if (view.controllerSessionOwnedByPreviousMount || !view.viewController.isPresented) {
           continue;
         }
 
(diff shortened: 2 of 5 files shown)


lodev09:
@maxlapides I've read changes here and I'm not confident about the changes. Do we still need this PR? Can you create a repro that I can actually replicate on my end via simulator? There's already a stress test in the example and is working fine. if you could modify it as a repro, that would be great.

lodev09:
Thanks for the thorough investigation here and in #793 — the repro made this easy to pin down.

I ported the repro into the example with the mid-dismissal remount shape and confirmed the event leak on `main`: every cycle delivered the previous sheet's `didDismiss` to the new mount, which tore down its lazy content and presented blank. That's the one state where the recycled host actually misbehaves on v4 — the auto-present wedge from v3 doesn't occur on `main` (the `finalizeUpdates` retry already covers it), and the navigation-dismiss paths were clean.

I went with a narrower fix in #818: detach the controller's delegate in `prepareForRecycle` only while it's mid-dismissal, retry auto-present once the presenter's dismissing sheet finishes, and make the module registry weak (your registered-tag finding — RN zeroes the tag before `prepareForRecycle`, so hosts were never unregistered). Everything else in the lifecycle stays as is, which keeps the pop/modal flows untouched.

Closing this in favor of #818. Appreciate the work.

maxlapides:
Awesome thanks for looking into this @lodev09! Sorry I missed your previous comment on this thread.

PR #804, opened 2026-08-26
fix(ios): Measure sheet content before presentation

maxlapides:
## Summary

Stage mounted sheet content in an invisible on-window host before presentation, then premeasure it inside a disposable child view controller before UIKit resolves the initial `auto` or `peek` detent.

Some native-backed children, including SwiftUI controls, change intrinsic size when they first acquire a containing view controller. Expo's SwiftUI host performs that reparenting from `didMoveToWindow`. Moving sheet content directly between two views that are already in the same window skips that callback, leaving the pre-presentation measurement stale. This change first moves content into the disposable controller while it is off-window, then attaches that controller to the hidden host. Nested hosting controllers therefore settle their intrinsic size before the opening animation captures its target detent.

Fixed detents keep their existing next-turn timing. Pending measurement work is still bounded by the existing 500 ms deadline and invalidated when presentation is superseded, content unmounts, or the component is recycled or deallocated.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Run `yarn test` (69 tests).
- Run `yarn typecheck`.
- Run `yarn prepare`.
- Run `xcrun clang-format --dry-run --Werror ios/TrueSheetView.mm`.
- Run `git diff --check`.
- Compile, install, and launch the patched package through an Expo SDK 57 / React Native 0.86 iOS app.
- On an iPhone 17 Pro simulator running iOS 26.5, verify both affected paths:
  - A lazy auto-height graphical SwiftUI date picker presents the complete six-week calendar in one continuous opening.
  - The biometric enrollment sheet reached through `/security-enroll-prompt` presents its complete content in one continuous opening.
- Compare the recordings with a deterministic frame analyzer. The reported date-sheet capture rises only 10 px over six frames with 69 px still remaining, then rises another 64 px. Two independent clean-build date-sheet presentations and the biometric enrollment presentation show no slow near-target interval followed by a second rise.

## Screenshots / Videos

The affected presentation paths were validated frame-by-frame. The consuming app is private, so its recordings are omitted here.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### 🐛 Bug fixes
 
+- **iOS**: Auto detents now measure mounted content before presentation, preventing late-measuring native views from resizing the sheet during its opening animation. ([#804](https://github.com/lodev09/react-native-true-sheet/pull/804) by [@maxlapides](https://github.com/maxlapides))
 - **Android**: A stacked parent sheet now follows its child frame-by-frame when the child resizes, and no longer gets stuck at stale detents after a half-expanded settle. ([#797](https://github.com/lodev09/react-native-true-sheet/pull/797) by [@lodev09](https://github.com/lodev09))
 - **Android**: The scrollable content now tracks the keyboard frame-by-frame — the focused input rides the keyboard up smoothly, and the scroll position eases back on dismiss instead of snapping. ([#798](https://github.com/lodev09/react-native-true-sheet/pull/798) by [@lodev09](https://github.com/lodev09))
 - The keyboard-driven scroll inset now accounts for a floating footer's height. ([#752](https://github.com/lodev09/react-native-true-sheet/pull/752) by [@lodev09](https://github.com/lodev09))
--- a/ios/TrueSheetView.mm
+++ b/ios/TrueSheetView.mm
@@ -50,6 +50,7 @@ static const NSTimeInterval InitialMeasurementTimeout = 0.5;
   RNScreensEventObserverDelegate>
 
 - (void)stageContainerViewForMeasurement;
+- (BOOL)stageContainerViewInMeasurementController;
 - (void)attachContainerViewToController;
 - (void)removeMeasurementHostView;
 - (void)scheduleInitialPresentationAfterMeasurement;
@@ -80,6 +81,7 @@ static const NSTimeInterval InitialMeasurementTimeout = 0.5;
   NSUInteger _initialMeasurementGeneration;
   CFTimeInterval _initialMeasurementDeadline;
   UIView *_measurementHostView;
+  UIViewController *_measurementViewController;
   NSArray *_pendingDetents;
   RNScreensEventObserver *_screensEventObserver;
 }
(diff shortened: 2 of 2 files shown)


lodev09:
@maxlapides do you have a repro for this fix? I want to test what we're fixing here. Thank you!

maxlapides:
Repro: https://github.com/maxlapides/true-sheet-swiftui-auto-detent-repro

Stock Expo 57 / RN 0.86.2 app — a button on a plain screen, and a sheet mounted when you tap it. The sheet is `detents={['auto']}` with `initialDetentIndex={0}`, and its content is an `@expo/ui` SwiftUI graphical `DatePicker` (pinned to August 2026 so it always renders a six-week grid).

On `4.0.0-beta.5` the sheet **opens twice**: it rises to the height of its header alone, sits there, then re-animates up to the height of the calendar. With this PR it opens once, continuously, to the same final height.

```bash
npm install && npm run ios       # 4.0.0-beta.5 — reproduces
npm run use:fix && npm run ios   # this PR — fixed
```

`use:fix` builds `refs/pull/804/head` from source and installs it; the rebuild is the point, since the change is native.

Rather than trust my eyes on a ~300 ms artifact, `scripts/analyze-opening.mjs` in the repo tracks the sheet's top edge per frame of a `simctl` recording. On an iPhone 17 Pro / iOS 26.5:

| | `4.0.0-beta.5` | this PR |
|---|---|---|
| first comes to rest at | `top=2378` | `top=1284` (final) |
| holds there for | 200–533 ms | never holds short |
| travel still remaining | **1092 px** | 0 px |
| separate animations | 2 | 1 |

Two runs of each variant, captured two independent ways — four for four, and every baseline run stalled at exactly the same place with exactly 1092 px left to go. I also verified it from a clean clone of the repo: plain `npm install`, `npx expo run:ios`, reproduces.

Frames sampled 133 ms apart, cropped to the sheet's travel range.

`4.0.0-beta.5` — parks as a header-only sliver, then jumps:



This PR — one continuous opening:



Happy to adjust the repro if you'd like a different shape of native-backed content in the sheet.

maxlapides:
Btw, this is a follow-up to the `lazy` feature that I was working on before. Unfortunately, `lazy` did not solve this issue like I thought it would. If you feel it's not valuable, we could also consider reverting the `lazy` feature.

lodev09:
@maxlapides I see. I did experience this when trying out expo/ui calendar so I can confirm. However, it looks like this isn't a real bug with true sheet. Feels like a workaround/hack to me.

For the `lazy` feature, it's a real feature that users can use like pre-rendering images like the `ScrollViewSheet` in example so that's fine.

I suggest fixing this in JS on your end, a fixed height maybe then cache it on subsequent present on the sheet. Or perhaps patch expo/ui calendar itself.

maxlapides:
@lodev09 Thanks for the recommendation, I'll try it out! Definitely don't want to add hacks to True Sheet :)

Issue #808, opened 2026-08-30
v4: setScrollableHandle: clears the scroll view but never re-resolves it


goranradulovichacs:
### Version

4.0.0-beta.6

### Platform

iOS 26.5, simulator (iPhone 17 Pro). React Native 0.86.2, New Architecture.

### What happens

`setScrollableHandle:` releases the resolved scroll view and relies on `setupScrollable` to find it
again, but nothing calls `setupScrollable` when the handle changes.

```objc
// ios/TrueSheetContentView.mm
- (void)setScrollableHandle:(NSInteger)scrollableHandle {
  if (_scrollableHandle == scrollableHandle) {
    return;
  }
  _scrollableHandle = scrollableHandle;

  // Release the previously resolved ScrollView — setupScrollable re-resolves
  [self clearScrollable];
}
```

`setupScrollable` runs only from lifecycle points — present, layout, size change,
`containerViewScrollViewDidChange`. A handle that arrives **after** the sheet has presented reaches
none of them. `_detectedScrollView` stays nil, so `applyContentInsetAdjustment` never runs and the
sheet never adopts the scroll view.

### When it shows

Whenever the scrollable mounts after presentation. In our app the sheet holds a sliding panel, and the
list lives in a panel step the reader opens later, so `scrollableRef` resolves well after
`onDidPresent`.

The list still scrolls, because it is a normal RN list, but nothing bounds it. Releasing a drag springs
it back to the top, which reads like a layout bug rather than a binding one.

### Measured

`NSLog` in `setScrollableHandle:` and in `findScrollView`, read through `simctl log stream`:

```
ACPROBE set  handle=7674 (was -1)
ACPROBE find handle=7674 view=nil
ACPROBE find handle=7674 view=nil
```

A walk of the whole `TrueSheetContentView` subtree at that moment found 78 views and no scroll view —
the handle arrives before Fabric inserts the view. Every later `find` also failed, because
`setupScrollable` was never called again to store it.

### Fix

```diff
   // Release the previously resolved ScrollView — setupScrollable re-resolves
   [self clearScrollable];
+
+  // A handle that arrives after the sheet has presented reaches no lifecycle call.
+  [self setupScrollable];
 }
```

With that one line, `find` returns `RCTScrollViewComponentView` and the sheet adopts the list.

`setupScrollable` already returns early when it finds nothing, so an early call costs nothing and the
existing lifecycle calls still retry.

### Note for consumers

[...]

PR #809, opened 2026-08-31
fix(ios): resolve the scrollable when its handle changes

godza:
### Summary

`setScrollableHandle:` releases the resolved scroll view and leaves the search to `setupScrollable`, as
its comment says. Nothing calls `setupScrollable` from there.

`setupScrollable` runs only from lifecycle points — present, layout, size change, and
`containerViewScrollViewDidChange`. A handle that arrives **after** the sheet has presented reaches none
of them, so `_detectedScrollView` stays nil, `applyContentInsetAdjustment` never runs, and the sheet
never adopts the list.

This one line resolves it where the handle changes.

### When it shows

Whenever the scrollable mounts after presentation. In my app the sheet holds a sliding panel and the list
lives in a step the reader opens later, so `scrollableRef` resolves well after `onDidPresent`.

The list still scrolls, because it is a normal RN list, but nothing bounds it. Fifteen rows with ten
visible reached row 13 and no further — a drag revealed the last two and releasing returned to 13.

### Measured

iPhone 17 Pro simulator, iOS 26.5, React Native 0.86.2, New Architecture, `4.0.0-beta.6`, with an
`NSLog` in `setupScrollable` read back through `simctl log stream`:

```
before   scrollH=649  contentH=821  adjustedContentInset.bottom=0
after    scrollH=676  contentH=821  adjustedContentInset.bottom=34
```

The inset is the proof: `applyContentInsetAdjustment` runs only once `_detectedScrollView` is set.

### Why it is safe

`setupScrollable` returns early when it finds nothing, so an early call costs nothing, and every existing
lifecycle call still retries.

One caveat worth stating: with this fix alone the resolve can still run before Fabric has inserted the
view, because the ref callback fires in the same commit. A consumer that reports its node a frame later
is picked up by `finalizeUpdates:` regardless. Both paths work; this makes the setter honest about what
its own comment promises.

Reported in #808.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - **Android**: The scrollable content now tracks the keyboard frame-by-frame — the focused input rides the keyboard up smoothly, and the scroll position eases back on dismiss instead of snapping. ([#798](https://github.com/lodev09/react-native-true-sheet/pull/798) by [@lodev09](https://github.com/lodev09))
 - The keyboard-driven scroll inset now accounts for a floating footer's height. ([#752](https://github.com/lodev09/react-native-true-sheet/pull/752) by [@lodev09](https://github.com/lodev09))
 - Auto detents now correctly size explicit-height and deeply nested scrollables. ([#783](https://github.com/lodev09/react-native-true-sheet/pull/783) by [@lodev09](https://github.com/lodev09))
+- **iOS**: A scrollable that mounts after the sheet has presented is now resolved — `setScrollableHandle:` released the previous scroll view without re-resolving, so the sheet never adopted the new one. ([#809](https://github.com/lodev09/react-native-true-sheet/pull/809) by [@godza](https://github.com/godza))
 
 ### 💥 Breaking changes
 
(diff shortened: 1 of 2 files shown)


lodev09:
Thanks for the PR and the thorough investigation in #808 — the bug is real, but tracing it showed this patch can't be the fix.

`setScrollableHandle:` on the content view has exactly one caller: `TrueSheetContainerView`'s `setupScrollable`, which sets the handle and then calls `[_contentView setupScrollable]` two lines later. The call this PR adds runs at the same pre-insert moment and finds nothing — verified with probes: on the handle commit, both existing resolve attempts (`updateProps` and `finalizeUpdates`) already run and fail because prop updates process before mount instructions within the same Fabric transaction. What actually fixed your app was the consumer-side change you mentioned (reporting the node a frame later) — that lands the handle in a later commit, which the existing prop-update path resolves on unpatched main.

The real fix needs a retry *after* the transaction's mount phase — done in #810 via `RCTMountingTransactionObserving` / `mountingTransactionDidMount:`, which covers both orderings without any consumer-side delay. Full trace details in https://github.com/lodev09/react-native-true-sheet/issues/808#issuecomment-5476820484.

Closing in favor of #810 — appreciate the report and the measurements, they made the root cause easy to pin down.

PR #818, opened 2026-09-02
fix(ios): keep a recycled host's dismissing controller from reaching the next mount

### LINKED ISSUE #789: Fast dismiss then re-present strands a late-mounted Fabric subtree until the next React commit (views configured but never inserted)

lodev09:
## Summary

A fast dismiss → remount cycle (unmounting a `TrueSheet` while it's still animating out and mounting a new one in the same commit) reuses the pooled Fabric host — and its `UIViewController` — for the new mount. The new sheet then received the previous sheet's `onDidDismiss` (tearing down its lazy content, so it presented blank) and its `initialDetentIndex` auto-present was blocked by `isBeingDismissed` with nothing to retry it.

- `prepareForRecycle`: when the controller is mid-dismissal, detach its delegate until the transition ends so the old session's tail (`didBlur`, `didDismiss`, position ticks) can't reach the next mount, then reattach.
- `presentInitialIfNeeded`: wait on the presenter's dismissing sheet's transition coordinator and retry instead of bailing.
- `TrueSheetModule`: the view registry is now weak. RN zeroes `view.tag` before `prepareForRecycle`/`dealloc`, so hosts could never unregister themselves and were retained forever.

Narrow alternative to #800 / #793 — only the mid-dismissal recycle state changes behavior; every other lifecycle path is untouched.

Fixes #789 (the TrueSheet side — the remaining "SVG fills missing until next commit" symptom is upstream RN Fabric, per the reporter's investigation).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Ported the reporter's repro into the example (not committed) with three auto-cycle modes on a physical iPhone, RN 0.86:

- **Mid-dismiss remount** (key-swap 30–330 ms into the dismissal): before, every cycle logged a stale `didDismiss` on the new mount and the sheet presented blank. After, 40+ cycles with `open → didPresent → willDismiss → remount` and no leaked event.
- **Remount on `onDidDismiss`** and **content swap mid-dismissal**: clean before and after.
- Navigation regressions checked: Open Modal → Basic sheet → Navigate Test → back → Dismiss Modal (main sheet re-presents), Dismiss Modal with a sheet open (single combined animation), Test Stack.
- `yarn test` passes.

## Screenshots / Videos

N/A

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- **iOS**: Unmounting a sheet mid-dismissal and mounting a new one no longer leaks the old sheet's `onDidDismiss` into the new mount or blocks its auto-present. ([#818](https://github.com/lodev09/react-native-true-sheet/pull/818) by [@lodev09](https://github.com/lodev09))
 - **iOS**: Fractional detents now resolve against the app's window instead of the full screen, fixing full-height sheets in windowed iPad apps (Stage Manager, iPadOS 26 windowing), and re-resolve when the window is resized or rotated. ([#815](https://github.com/lodev09/react-native-true-sheet/pull/815) by [@vilindberg](https://github.com/vilindberg) and [@lodev09](https://github.com/lodev09))
 - **Android**: A stacked parent sheet now follows its child frame-by-frame when the child resizes, and no longer gets stuck at stale detents after a half-expanded settle. ([#797](https://github.com/lodev09/react-native-true-sheet/pull/797) by [@lodev09](https://github.com/lodev09))
 - **Android**: The scrollable content now tracks the keyboard frame-by-frame — the focused input rides the keyboard up smoothly, and the scroll position eases back on dismiss instead of snapping. ([#798](https://github.com/lodev09/react-native-true-sheet/pull/798) by [@lodev09](https://github.com/lodev09))
--- a/ios/TrueSheetModule.h
+++ b/ios/TrueSheetModule.h
@@ -27,15 +27,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  * Register a sheet component view with its React tag
- * Called automatically by TrueSheetView during initialization
+ * Called automatically by TrueSheetView when it attaches to a window.
+ * The registry holds views weakly, so no unregistration is needed.
  */
 + (void)registerView:(TrueSheetView *)view withTag:(NSNumber *)tag;
-
-/**
- * Unregister a sheet component view
- * Called automatically by TrueSheetView during dealloc
- */
-+ (void)unregisterViewWithTag:(NSNumber *)tag;
 
 @end
 
--- a/ios/TrueSheetModule.mm
+++ b/ios/TrueSheetModule.mm
@@ -18,8 +18,9 @@
 
 #import <TrueSheetSpec/TrueSheetSpec.h>
 
-// Static registry to store view references by tag
-static NSMutableDictionary<NSNumber *, TrueSheetView *> *viewRegistry;
+// Static registry of live hosts by tag — weak so an unmounted host deallocates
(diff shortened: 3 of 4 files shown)


mmichels-brex:
hey @lodev09! first of all, thank you so much for this library 🙏 
are you planning to ship this fix soon? it'd be greatly appreciated 🙌

lodev09:
@mmichels-brex released on v4 beta 8

PR #820, opened 2026-09-02
fix(android): skip explicit Kotlin plugin when AGP provides built-in Kotlin (v3)

lodev09:
## Summary

Port of #819 to v3.

AGP 9 ships built-in Kotlin and applies the Kotlin plugin itself. Applying `kotlin-android` again fails configuration with `Cannot add extension with name 'kotlin'`. The explicit apply is now skipped when AGP provides Kotlin: AGP 8 and older always apply, AGP 10+ never applies, AGP 9 applies only with `android.builtInKotlin=false`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- `android/build.gradle` is byte-identical to main after #819.
- On main, the bare example passed a clean Android build (AGP 8.12, RN 0.86).
- AGP 9 path verified by the original author on RN 0.87 / Expo SDK 58 with AGP 9.2.1.

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **Android**: The library now builds on Android Gradle Plugin 9, which ships built-in Kotlin — the explicit Kotlin plugin is skipped when AGP already provides it. ([#819](https://github.com/lodev09/react-native-true-sheet/pull/819) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+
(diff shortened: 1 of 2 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.13. See [Release 3.11.13](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.13) for release notes._

PR #823, opened 2026-09-04
fix(ios): let TrueSheetOverlay children install their own view controllers

SamuelBrucksch:
## Problem

Content that installs a `UIViewController` renders nothing inside `TrueSheetOverlay`, with no error or warning. The most common case is a React Navigation native stack (via `react-native-screens`): the navigator mounts, React renders it, and the overlay draws nothing. Plain views in the same overlay render fine.

```tsx
<TrueSheetOverlay>
  <View style={StyleSheet.absoluteFill}>
    <NativeStack.Navigator>
      <NativeStack.Screen name="A" component={A} />
    </NativeStack.Navigator>
  </View>
</TrueSheetOverlay>
```

Worth knowing when reproducing: **a hot reload hides the bug.** A controller that is already attached skips the code path below, so the overlay looks correct until the app is cold-started.

## Cause

`TrueSheetOverlayView` adds its container straight to the window:

```objc
_containerView.frame = window.bounds;
[window addSubview:_containerView];
```

That makes the container a *sibling* of `rootViewController.view`. A `UIWindow`'s responder chain does not include its root view controller — a view controller only joins the chain for views inside its own view — so `-[UIView reactViewController]`, which walks `nextResponder`, finds no view controller for anything inside the container.

`react-native-screens` adds its `UINavigationController`'s view only inside the branch guarded by that lookup:

```objc
// RNSScreenStack.mm
UIView *parentView = (UIView *)self.reactSuperview;
while (parentView) {
  if (parentView.reactViewController) {
    [parentView.reactViewController addChildViewController:controller];
    [self addSubview:controller.view];   // ← never reached in an overlay
    ...
```

With the lookup returning nil the loop ends having done nothing, so the navigation controller's view is never added. Hence: mounted, no error, nothing drawn.

Android does not have this gap — `TrueSheetOverlayRootView` implements `RootView`, so children have something to attach to. This PR is the iOS equivalent.

## Fix

Give the container a `UIViewController` owner. Assigning the container as that controller's `view` is what puts the controller in the responder chain, so `reactViewController` resolves and child view controllers attach as they do elsewhere in the tree. It is also added as a child of the window's root view controller so it takes part in the normal lifecycle.

No behaviour change for plain views, which need no parent view controller.

## Notes

Happy to reshape this — a few things are your call:

- Once a view controller owns that container, UIKit consults it for status-bar style, home-indicator behaviour and rotation. Overriding those to defer to the root view controller may be desirable.
- `addChildViewController:` affects appearance-callback ordering, and can be dropped if the responder chain alone is enough.
- If hosting view controllers is out of scope for the overlay, a documented note plus a dev-mode warning when the lookup fails would still be a big improvement over silent non-rendering. I see #822 documents this for *sheet screens*; the overlay case is separate and currently undocumented.

## Testing

Verified on the iOS 26 simulator, new architecture: a native stack inside `TrueSheetOverlay` renders and navigates correctly above a presented sheet, on a cold start. Plain overlay content (a fixed-size view) is unaffected. Not yet tested on iPad or in a windowed/Stage Manager configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### 🐛 Bug fixes
 
-- **iOS**: `TrueSheetOverlay` children can now install their own view controllers — its container is owned by a `UIViewController`, so `-[UIView reactViewController]` resolves inside the overlay as it does elsewhere in the tree. Previously such content mounted and rendered nothing. ([#823](https://github.com/lodev09/react-native-true-sheet/pull/823) by [@SamuelBrucksch](https://github.com/SamuelBrucksch))
+- **iOS**: Content that needs a parent view controller, like React Native's `Modal`, now works inside `TrueSheetOverlay` instead of silently rendering nothing. ([#823](https://github.com/lodev09/react-native-true-sheet/pull/823) by [@SamuelBrucksch](https://github.com/SamuelBrucksch) and [@lodev09](https://github.com/lodev09))
 - **iOS**: Unmounting a sheet mid-dismissal and mounting a new one no longer leaks the old sheet's `onDidDismiss` into the new mount or blocks its auto-present. ([#818](https://github.com/lodev09/react-native-true-sheet/pull/818) by [@lodev09](https://github.com/lodev09))
 - **iOS**: Fractional detents now resolve against the app's window instead of the full screen, fixing full-height sheets in windowed iPad apps (Stage Manager, iPadOS 26 windowing), and re-resolve when the window is resized or rotated. ([#815](https://github.com/lodev09/react-native-true-sheet/pull/815) by [@vilindberg](https://github.com/vilindberg) and [@lodev09](https://github.com/lodev09))
 - **Android**: A stacked parent sheet now follows its child frame-by-frame when the child resizes, and no longer gets stuck at stale detents after a half-expanded settle. ([#797](https://github.com/lodev09/react-native-true-sheet/pull/797) by [@lodev09](https://github.com/lodev09))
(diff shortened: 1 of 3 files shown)


lodev09:
Thanks for the thorough writeup. Verified locally and pushed a small commit to your branch (docs + changelog trim).

**Findings**

- Confirmed on iOS: `RNSScreenContainer` and RN `Modal` inside `TrueSheetOverlay` render nothing on `main` and work with this PR. `Modal` is the strongest reason to take this since it fails silently for a very ordinary use case.
- The Android claim isn't accurate: `ScreenContainer.setupFragmentManager` checks for the concrete `ReactRootView` class (Fabric's `ReactSurfaceView` extends it) or a react-native-screens `FragmentProviding` view, not the `RootView` interface. So `ScreenContainer`/native stacks inside the overlay crash on Android with the same `ScreenContainer is not attached under ReactRootView` error as sheet screens. RN `Modal` works on Android since it's dialog-based.
- Given that, react-native-screens containers inside the overlay stay unsupported on both platforms. I added a Limitations bullet in the overlay guide pointing to the existing note from #822 rather than treating stacks-in-overlay as a supported pattern.
- Changelog entry trimmed to the repo's one-liner style.

**On `addChildViewController:`**

I'd lean towards dropping the containment and keeping only the `view` assignment. Beyond simplicity, there's a presentation concern: a `Modal` inside the overlay presents from the container controller, and as a child of the root that resolves to the root as presenter. If a sheet is already presented from the root, UIKit should refuse with "already presenting". A standalone controller with its view in the window would present from itself. Could you check `Modal` from the overlay while a sheet is up, with and without the containment? If the standalone variant works there, that settles it.

lodev09 (approved):

PR #825, opened 2026-09-05
fix(web): stacked sheet dismissal, peek resize, and cascade fixes (v3)

### LINKED ISSUE #816: Web: a press inside a stacked sheet dismisses the sheets beneath it; the same press's release cancels the dismiss animation
Affects the web implementation, verified on 3.11.9 through 3.11.12.

## Bug 1 — press inside a stacked sheet dismisses its parents

On web every sheet is its own Radix dialog whose wrapper portals as a sibling into the shared container. Radix's `pointerDownOutside` only treats a sheet's own content as "inside", so a pointer down inside a stacked descendant sheet is delivered as an outside press to every sheet beneath it — the parent calls `closeDrawer` and dismisses itself. With a navigation stack of sheets, any tap inside the top sheet (a header back button, a footer button, even inert content) tears down the sheets below it.

`handlePointerDownOutside` in `TrueSheet.web.tsx` already guards the portal-container and detached-footer cases; it needs a third guard: ignore presses whose target sits inside a wrapper that comes later in document order (presentation order) than the sheet's own wrapper. Document order rather than the `useSheetStack` registry matters because Radix can dispatch the handler on the deferred click event, after a dismissing descendant has already left the stack; treating `DOCUMENT_POSITION_DISCONNECTED` as "above" keeps the failure mode safe (a suppressed dismissal rather than a spurious one).

## Bug 2 — pointer release snaps a dismissing sheet back, cancelling its close animation

A press marks `isDragging` on pointer down, so the release of the same press that dismissed a sheet (e.g. a back button that pops it) still runs `release()` → `onReleaseSnapPoints` → `snapToPoint`, which unconditionally resets the wrapper transform — undoing the slide-out the dismiss effect just queued. The sheet freezes in place and blinks out on unmount instead of animating. Guarding `release()` with `if (!isOpen) return` (after the drag-state cleanup, still firing `onReleaseProp` so consumer drag state resets) fixes it; drag-to-dismiss is unaffected since its release runs while the drawer is still open.

## Current workaround

We ship both guards as a postinstall patch in input-output-hk/lace-platform (PR #2680) and would be happy to upstream them as a PR if the approach sounds right.

lodev09:
## Summary

Backport of #824 to v3. Cherry-picked with `-x`; the only adaptations are v3's element casts in place of `getDOMElement`, the layout hook on v3's own content views, and the geometry helper returning the effective height and ceiling.

- A press inside a stacked sheet (its content, footer, or dim overlay) no longer dismisses the sheets beneath it.
- The release of a press that already dismissed the drawer no longer snaps it back and cancels the close animation.
- Dismissing a sheet dismisses every sheet stacked above it, matching iOS and Android.
- The peek detent re-measures when content above the `TrueSheetPeek` marker changes.
- Form sheets cascade behind a child like page sheets; the parent stays put only at full height (detent 1).
- A grandparent follows its child's actual pushed position. Parents stay put while the child is dragging.

Closes #816

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Driven in the browser with Playwright against the v3 example app:

- Sheet navigator: press inside Settings stacked over Details, both stay; dim click dismisses Settings only, then Details; hold pointer, Escape, release keeps the slide-out transform.
- Test screen: Add/Remove Content grows and shrinks the visible peek; Basic slides to the child's top at Peek and stays at full height; parent Dismiss takes the child.
- Map screen: main follows Basic's position as Basic resizes and while Basic is stacked under its child.

## Screenshots / Videos

n/a

## Checklist

- [ ] I tested on iOS
- [ ] I tested on Android
- [x] I tested on Web


lodev09:
:rocket: _This pull request is included in v3.11.13. See [Release 3.11.13](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.13) for release notes._

PR #829, opened 2026-09-08
fix(ios): let VoiceOver dismiss the sheet and reach the grabber (v3)

lodev09:
## Summary

Backport of #828 to v3. Fixes #807.

Under VoiceOver, iOS sheets could not be dismissed: the two-finger Z (escape) gesture was ignored, the grabber was not in the accessibility tree, and the dim view was excluded from traversal.

- The sheet's root view is now a `TrueSheetControllerView` that handles `accessibilityPerformEscape`, dismissing when `dismissible` is true.
- The grabber is exposed as the first accessibility element. With `grabberOptions` that is the library grabber (swipe down at the lowest detent dismisses). Without it, UIKit's system grabber is picked up from the presented view, skipping the transparent bottom duplicate UIKit also adds.
- Removed the container's `accessibilityViewIsModal`, which filtered out the grabber sibling. The root view already carries the modal flag.
- `setupGrabber` refreshes the exposed elements for the topmost sheet only, so a parent's prop update cannot steal the window override from a child.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Clean cherry-pick of the main commit; the touched code is identical between branches. Verified on main (#828) by building the example on the iOS 26.5 simulator, inspecting the accessibility tree, and calling `accessibilityPerformEscape` via lldb across dimmed, non-dimmed, non-dismissible, custom grabber, system grabber, and stacked cases. Not rebuilt on v3.

## Screenshots / Videos

N/A

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🐛 Bug fixes
+
+- **iOS**: VoiceOver users can now dismiss the sheet with the two-finger Z gesture, and the grabber is announced as the sheet's first element so it can be resized or dismissed with swipe gestures. ([#829](https://github.com/lodev09/react-native-true-sheet/pull/829) by [@lodev09](https://github.com/lodev09))
+
 ## 3.11.13
 
 ### 🐛 Bug fixes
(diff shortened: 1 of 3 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.14. See [Release 3.11.14](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.14) for release notes._

PR #831, opened 2026-09-09
fix(ios): escape from the system grabber, keep custom grabber from expanding under VoiceOver (v3)

lodev09:
## Summary

Port of #830 to v3. Follow-up to #829 from device testing in #807.

- The two-finger Z escape did nothing while VoiceOver focus was on UIKit's system grabber. VoiceOver walks superviews from the focused element, and the system grabber is a sibling of our root view, so the walk went presented view → window and never reached the escape handler. The presented view now handles escape via `accessibilityPerformEscapeBlock` (iOS 17+), forwarding to the controller.
- With a custom grabber, presenting a sheet under VoiceOver expanded it to the next detent on its own. The grabber's hit-padded frame started above the sheet's top edge, and UIKit reveals a clipped focused element in a sheet by changing detent. The frame is now clamped inside the sheet.
- The custom grabber's accessibility frame is padded evenly around the pill, matching the system grabber's focus box.

Tapping the dimmed background is not a VoiceOver target, matching native UIKit sheets.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Same changes verified on main (#830) with the bare example on a physical device with VoiceOver on:

- Z escape from the system grabber and custom grabber dismisses.
- Non-dismissible sheet: escape plays the failure sound and stays.
- Stacked: escape dismisses only the child, focus returns to the parent.
- Custom grabber sheet presents at its initial detent and stays there. Grabber double tap and swipe up/down still cycle detents, swipe down at the lowest detent dismisses.
- Focus box on the custom grabber matches the system grabber.

## Screenshots / Videos

N/A

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🐛 Bug fixes
 
-- **iOS**: VoiceOver users can now dismiss the sheet with the two-finger Z gesture, and the grabber is announced as the sheet's first element so it can be resized or dismissed with swipe gestures. ([#829](https://github.com/lodev09/react-native-true-sheet/pull/829) by [@lodev09](https://github.com/lodev09))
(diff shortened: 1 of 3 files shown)


lodev09:
:rocket: _This pull request is included in v3.11.14. See [Release 3.11.14](https://github.com/lodev09/react-native-true-sheet/releases/tag/v3.11.14) for release notes._

PR #840, opened 2026-09-14
fix(ios): ignore pan gestures that start outside the sheet

SamuelBrucksch:
## Summary

A sheet can be dragged by a gesture that never touched it.

With interactive content behind an undimmed sheet — a map, for example — press down on that content, pan, and then keep dragging onto the sheet without lifting. The sheet starts following the finger and settles at another detent when it is released, while the content behind stops tracking mid-gesture.

The recognizer responsible is not on the presented view. UIKit presents a sheet inside container views that span the screen, and a pan recognizer installed on one of those receives touches that began anywhere, including outside the sheet.

That is also why `draggable` does not prevent it: `setPanGesturesEnabled:forView:` walks only `presentedView.gestureRecognizers`, so the recognizers on the containers above it stay enabled. Setting `draggable={false}` and `scrollableOptions.scrollingExpandsSheet = false` together still leaves the sheet draggable this way.

### The fix

`restrictPanGesturesToSheetView:` walks from the presented view up to the window and gives each pan recognizer a delegate that returns `NO` from `gestureRecognizer:shouldReceiveTouch:` when the touch's initial location falls outside the presented view's bounds. Gestures that begin on the sheet are unaffected.

Two details worth noting for review:

- **The existing delegate is preserved.** The shim keeps a weak reference to whatever delegate UIKit had installed, forwards `shouldReceiveTouch:` to it when the touch is inside the sheet, and forwards every other callback through `respondsToSelector:` / `forwardingTargetForSelector:`. Replacing the delegate outright would change presentation behaviour.
- **Scroll views are skipped.** `-[UIScrollViewPanGestureRecognizer setDelegate:]` raises rather than accepting a replacement — a scroll view is its own pan recognizer's delegate and depends on those callbacks. Including them crashes on the first presentation, so the walk skips a scroll view's built-in pan.

The filter is applied in `setupGestureRecognizer`, which already runs on presentation and prop updates, and is idempotent: an existing filter has its `boundsView` refreshed rather than being wrapped in another one. It is retained via an associated object because `UIGestureRecognizer.delegate` is weak.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Tested on the iOS Simulator (iPhone 17, iOS 26.5) with a full-screen map behind a sheet using `['peek', 'auto']` detents and an undimmed collapsed detent.

Before the change, pressing on the map, panning, and continuing onto the sheet drags the sheet to another detent. After it, the map keeps tracking the gesture across the sheet and the sheet stays put.

Checked for regressions in the interactions this touches:

- Dragging the sheet from its own surface still works in both directions and settles at the expected detents.
- Content inside the sheet still scrolls to the end.
- A second sheet still presents over the first and swipe-to-dismiss still works.

Not tested on Android or Web — the change is iOS-only.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web


lodev09:
@SamuelBrucksch I think this is native behavior on ios -- most likely for accessibility reasons. Can you try the native Maps app on ios? See if you could replicate the same issue.

SamuelBrucksch:
Indeed, this is the same on Apple Maps, the sheet moves down when the map is dragged... Its a bit annoying though in my opinion. Should I hide it between a property which needs opt in?

lodev09:
@SamuelBrucksch yeah I figured. I'm staying true to native behavior as much as possible so I'd rather leave it as-is. Feel free to patch your own if it fixes your issue.

Thanks for your contribution! Closing this.

PR #843, opened 2026-09-15
fix(example): adopt UIScene lifecycle for Xcode 27

lodev09:
## Summary

The bare example crashed at launch on the iOS 27 simulator with Xcode 27:

```
Application failed to launch: UIScene life cycle is required for apps built with this SDK.
```

Apps built against the iOS 26 SDK got a runtime *warning* for not adopting the UIScene lifecycle. The **iOS 27 SDK makes it fatal** — UIKit refuses to launch. The example still used the classic `UIApplicationDelegate` + `UIWindow(frame:)` setup with no scene manifest.

This is not a TrueSheet bug — the library's iOS code needed no changes. It's the example app's native lifecycle.

### Changes

- **`SceneDelegate.swift`** (new) — creates the `UIWindow` from the `UIWindowScene` and calls `startReactNative` in `scene(_:willConnectTo:options:)`.
- **`AppDelegate.swift`** — no longer creates the window or starts RN; it only builds and retains the RN factory.
- **`Info.plist`** — adds `UIApplicationSceneManifest` pointing at `$(PRODUCT_MODULE_NAME).SceneDelegate`.
- **`project.pbxproj`** — registers the new file.

### Notes

**A scene manifest alone is not a fix.** It silently converts the crash into a blank screen, because the window built in `AppDelegate` is never attached to the `UIWindowScene`. I verified this by patching the installed bundle's plist. The window has to be created *from* the scene. Same trap documented in [expo/expo#47570](https://github.com/expo/expo/issues/47570).

`AppDelegate` keeps a **weak** `window` property on purpose. RN still reads `delegate.window` in `RCTLogBoxView.dealloc` ([react/react-native#56571](https://github.com/react/react-native/issues/56571), still present in RN 0.86 at `React/CoreModules/RCTLogBoxView.mm:87`). Since `window` is an *optional* protocol property, dropping it would cause an unrecognized-selector crash whenever LogBox is dismissed in dev. Weak because `SceneDelegate` owns the window.

`launchOptions` is now `nil` — under the scene lifecycle those arrive via `connectionOptions`. The example doesn't read them, so no behavior change.

### Upstream status

- **React Native**: no fix merged. [#54763](https://github.com/react/react-native/pull/54763) and RFC [#967](https://github.com/react-native-community/discussions-and-proposals/pull/967) are both still open. Hand-rolling the SceneDelegate is currently the only option for bare RN.
- **Expo**: fixed in **57.0.23** ([expo/expo#50191](https://github.com/expo/expo/pull/50191)), plus an opt-in config plugin `@config-plugins/expo-uiscene-lifecycle` that is merged but not yet published to npm.

The **Expo example is still affected** and is not covered by this PR — it pins `~57.0.7` and its `ios/` is prebuild-generated. Worth a follow-up once the plugin publishes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- `xcodebuild` → **BUILD SUCCEEDED** against the iOS 27 SDK.
- Installed and launched on the iOS 27.0 simulator (iPhone 18 Pro Max): zero `NoSceneLifecycleAdoption` errors, process stays alive, app renders fully — map, sheet, grabber, header/footer, liquid glass blur.
- Confirmed the `window`/`setWindow:` selectors are still exposed to the ObjC runtime by dumping ObjC class metadata from the built binary (Swift→ObjC thunks present), so the LogBox path is safe.

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android
- [ ] I tested on Web


alex-strae:
You saved my sanity

PR #851, opened 2026-09-17
fix(android): keep the peek fold when screens swap

SamuelBrucksch:
## Summary

`attachPeekView` rejects a second peek and returns. React Navigation mounts the incoming screen's peek before it unmounts the outgoing one's, so during a swap between two screens that each render a `TrueSheetPeek`, the new peek is refused and the old one then detaches — leaving the sheet with no peek at all. Its peek detent collapses to about header height, so navigating back to a peek screen drops the sheet, showing only its top edge.

Rendering the marker only while a screen is focused does not avoid this: both screens are briefly focused across the transition.

The last-attached peek now wins. The previous peek's delegate is cleared immediately, and `detachPeekView`'s `peekView !== view` guard already makes its later detach a no-op, so the incoming peek survives.

The warning is kept for genuine misuse — two peeks live in one sheet — but deferred by a frame, since a swap's outgoing peek is no longer attached to a window by then. That also removes a misleading warning on every such navigation.

iOS has the same guard in `attachPeekView` and I have left it alone: I could not reproduce the collapse there, so the ordering appears to differ. Happy to mirror it if you would prefer symmetry.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

Two app-sheet screens, each rendering a `TrueSheetPeek`, in a React Navigation stack with detents `[fraction, 'peek', 'auto']`. Navigating forward, then back:

- before: the sheet drops to roughly header height and its content is unreachable, with `Sheet can only have one peek component.` logged on each transition
- after: the peek fold is intact — header, rows and footer all present — and nothing is logged

Verified on an Android 16 emulator (Pixel 8) against 4.0.0-beta.7, where `attachPeekView` is identical to `main`.

## Screenshots / Videos

N/A

## Checklist

- [x] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web


lodev09 (dismissed):
Thanks @SamuelBrucksch

PR #857, opened 2026-09-18
fix(android): rest a two-detent sheet at its expanded stop

SamuelBrucksch:
## Summary

A scrollable inside a sheet with **two** detents could not be scrolled while the sheet sat
at its top detent. The drag did nothing: the list stayed put and the sheet, already fully
open, had nowhere to move.

`halfExpandedRatio` was derived from the *top* detent, so the half stop landed a few pixels
below the expanded stop — both resolve to the largest available height. Dragging up from
the expanded stop then satisfies `newTop >= getExpandedOffset()` in
`BottomSheetBehavior.onNestedPreScroll`, so the behavior consumes the whole delta
(`consumed[1] = dy`) to cross that gap instead of passing it to the scrollable, and on
release settles back — the state flips `STATE_EXPANDED` → `STATE_HALF_EXPANDED` on every
attempt and the scroll position never changes.

The half stop now shares the peek height for two detents, exactly as it already does for
one, where the existing comment notes the same reasoning. The state map follows, mapping
`STATE_HALF_EXPANDED` to index 0 alongside `STATE_COLLAPSED`. A keyboard floor keeps its
distinct half stop, since it rests there.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

A two-detent sheet whose content is a list, with `scrollableRef` pointing at it:

```tsx
<TrueSheet ref={sheet} detents={['peek', 1]} scrollableRef={list} style={{ flex: 1 }}>
  <FlatList ref={list} data={data} renderItem={renderItem} />
</TrueSheet>
```

Before, at the top detent: dragging up on the list does nothing.
`setOnScrollChangeListener` never fires, and the behavior's state reads
`STATE_HALF_EXPANDED` (6) after the first drag.

After: the list scrolls in both directions from the first touch, the state stays
`STATE_EXPANDED` (3), and dragging down at the top of the list still collapses the sheet to
the peek.

Verified on an Android 17 emulator against `4.0.0-beta.15`, instrumented to log the
behavior state and every scroll position change: from no scroll events at all to ~350
across the full content height. Three-detent and single-detent sheets are unaffected — only
the two-detent branch changes — and this is Android-only, so iOS and Web are untouched.

## Checklist

- [ ] I tested on iOS
- [x] I tested on Android
- [ ] I tested on Web


SamuelBrucksch:
Closing this — I could not sustain the claim.

The symptom I opened it for (a scrollable not scrolling at a two-detent sheet's top detent) turned out to be caused by something in my own app, not by the detent mapping: a horizontal `ScrollView` above the list was passing `nestedScrollEnabled`, so `BottomSheetBehavior.findScrollingChild` took that row as its nested-scrolling child. Every vertical drag then counted as outside the scrolling child and was intercepted into a sheet already at its largest detent. Removing the prop fixed it.

I re-tested against a pristine 4.0.0-beta.15 with only that app-side change, and the list scrolls normally — so the detent change here was not load-bearing and I would rather not ship a fix whose justification does not hold.

One observation that may still be worth something, offered as-is: with two detents the sheet does settle in `STATE_HALF_EXPANDED` at its largest detent, since `halfExpandedRatio` is derived from `detents[1]` and both stops resolve to the largest available height. I saw the state flip `STATE_EXPANDED` -> `STATE_HALF_EXPANDED` on each upward drag. It caused no visible problem once the real cause was fixed, so it may well be harmless.

Separately, this sentence in the scrolling guide cost me a while: "any `nestedScrollEnabled` prop on your `ScrollView` or `FlatList` is ignored". That is true for the scrollable passed to `scrollableRef`, which the library sets itself, but on any other scrollable in the sheet content the prop still applies — and enabling it there hands the behavior the wrong scrolling child. Happy to open a small docs PR clarifying that if useful.

lodev09:
@SamuelBrucksch feel free to open up PR for the docs related to the ScrollView `nestedScrollEnabled`. Also, you can play around with the example app to try to reproduce your issue -- `ScrollViewSheet` has that setup as well (horizontal scroll chips).

PR #859, opened 2026-09-19
fix(ios): dismiss a sheet that is unmounted while still presented

SamuelBrucksch:
## Summary

On iOS, a `TrueSheet` that is unmounted while still presented stays on screen: fully drawn, dead to touches, and gone from the accessibility tree, until the user swipes it away. Only a mid-dismissal controller was handled in `prepareForRecycle`; a presented one was left alone. `dealloc` would dismiss it, but Fabric pools the host view, so `dealloc` never runs in practice. Android already dismisses in `onDropInstance`.

This dismisses the controller in `prepareForRecycle` when it is still presented (nil-ing the delegate first, so no events reach the unmounted JS side), and lets the existing mid-dismissal block take it from there.

Minimal repro on `main`:

```tsx
const [show, setShow] = useState(true)

<Button title="Unmount" onPress={() => setShow(false)} />
{show && (
  <TrueSheet initialDetentIndex={0} detents={['auto']}>
    <Text>Unmount me without dismissing</Text>
  </TrueSheet>
)}
```

Before: the sheet remains after "Unmount"; nothing behind it can be tapped. After: it animates out like a `dismiss()`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Test Plan

- Unmounted a presented sheet by conditionally rendering it (as above) — the sheet dismisses, the screen underneath is interactive again, and a fresh mount presents normally.
- Unmounted a sheet that was presented over another sheet — the parent regains focus (`emitDidDismissEvents` still runs in `viewDidDisappear`, so `_isPresented` resets and the parent's focus callbacks fire).
- Regression: `dismiss()`, swipe-to-dismiss, dim tap, and unmounting *mid-dismissal* (the existing `isBeingDismissed` path) behave as before; no `TrueSheet:` warnings in the system log.

## Screenshots / Videos

n/a

## Checklist

- [x] I tested on iOS
- [ ] I tested on Android (no Android change)
- [ ] I tested on Web (no Web change)


lodev09:
This is breaking with react-navigation screens + undimmed sheets. When you dismiss a screen with a presented sheet, ios waits for the sheet animation to finish before the screen dismisses causing bad UX. You can compare behavior in the example app.

